Issue #11, opened 2020-08-01
Simple Main With Test


alyhasansakr:
Write a simple main function and a corresponding test function, a function that returns Hello World is enough.
Then create an action in GitHub to do the test automatically.

juan-castrillon:
When reading about Github actions, I come across the option to publish the resulting built packages into GitHub Packages or Maven Central Repository (see [here](https://docs.github.com/en/actions/publishing-packages-with-github-actions/about-packaging-with-github-actions)). I would say, is not necessary right now, but what do you think?

alyhasansakr:
Not important for the infrastructure manager, we are only interested in tests

alyhasansakr:
We should be able to see test results when a pull-request is made here, @juan-castrillon go ahead and make one with a test.

PR #12, opened 2020-08-03
Create yml file 

juan-castrillon:
### Created the YML file used to trigger and control Github actions.

How the different commands work is explained in the comments of the file, 

However, after running some test in a repository of my own, with a simple app the test took ~40s. Its important to remember that [we only have 2000 minutes of test a month](https://docs.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions#about-billing-for-github-actions).

In order to make the execution time faster, I added the cache action which is supposed to help. If we think that we need it to be even faster, that could be a topic for future work . (Evaluate the JDK setting, and the gradle running, which takes all of the time)

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -1,0 +1,31 @@
+name: Automatic Testing - Gradle
+
+#The action will be triggered everytime a pull request is made that includes files in the src directory (master branch)
+on: 
+  pull_request:
+    branches: 
+      - master
+    paths:
+      - 'src/**'
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2 ##Copies the repository files in the remote machine
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1 #Sets up the Java version (8)
+      with:
+        java-version: 1.8
+    - name: Cache Gradle packages #Saves gradles dependencies after a succesful run, so they dont have to be downloaded again, if there are changes in the build script skips the cache
+      uses: actions/cache@v2
+      with:
+        path: ~/.gradle/caches
+        key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+        restore-keys: ${{ runner.os }}-gradle
+    - name: Grant execute permission for gradlew
+      run: chmod +x gradlew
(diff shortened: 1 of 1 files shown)


juan-castrillon:
Also, as soon as this gets merged, I will submit the request that closes #11 , and you guys will be able to see the test being performed on the new code

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,31 @@
...
+  pull_request:
+    branches: 
+      - master
+    paths:
+      - 'src/**'
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
```
alyhasansakr, line 14:
```suggestion
    runs-on: windows-latest
```
We are mainly using windows now, later we are going to add Linux as well

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,31 @@
...
+
+    steps:
+    - uses: actions/checkout@v2 ##Copies the repository files in the remote machine
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1 #Sets up the Java version (8)
+      with:
+        java-version: 1.8
```
alyhasansakr, line 21: any specific reason we should use this version of java?

.github/workflows/gradle.yml
juan-castrillon: Yes, but this is only the GitHub hosted virtual machine in which the action gets executed. (I mean it doesn't influence our development at all) so  I thought of going with the default configuration. Do we really need to differentiate here?

.github/workflows/gradle.yml
juan-castrillon: Not really. I just went with default again. Is there any specific version of Java we should define? I tought we could keep this one, and then modify it if we need any functionality of a higher version. But do you think we should configure it to any specific one?

.github/workflows/gradle.yml
alyhasansakr: normally not, but if we do something platform dependent (like network sockets), then we will need to differentiate. Anyway we can keep it as it is now and see if it causes problems in the future.

.github/workflows/gradle.yml
alyhasansakr: ok, lets just default for now.

PR #13, opened 2020-08-05
Automatic testing

juan-castrillon:
A gradle build file was added, however, some of the jar functionality is not used because right now the test doesnt build but only tests. 

The main java file just outputs "Hello World!".  And I implemented two tests, to check if the output is null and if the output is exactly how it should be

As future improvements, the testing task could be specified in the build script to output more details into console in case a test fails.


juan-castrillon:
Closes #11

juan-castrillon:
When you say build files are you talking about the whole build folder? Because inside there is some things generated by the test task as well (Test HTML page showing the results and reports) so im thinking in leaving these. What do you think?

alyhasansakr:
yes the whole build folder, actions have their own cache no need to keep reports or anything in the remote repo.

PR #16, opened 2020-08-07
Master class v1

### LINKED ISSUE #15: The Master Class
The Master class is the main entry point of our back-end.
The Master has two streams; input and output. For example, providing "deploy application" in the input gives "Helm chart execution" output, here the master should not execute the Helm chart on its own, instead, it just sends a message for Helm controller (another application).
The Master also has the configuration of the platform, which should be written in an extensible format (JSON/XML) and read from a file.
The Master is our default place for any function that we don't know yet where it should be. We have to refactor it regulatory to make sure it doesn't get out of control.

As we still don't know how most of the other functionalities look like, what is required here is to just write a simple class that can do the functionality mentioned above (output is just printing for now), including tests of course.
Create other classes as necessary, you can look at the class diagram in #3 but you can also just write your own.

This is our first piece of code, so there will be a lot of modifications and discussions. We still don't have code rules, we are also going to develop that with this ticket. @theodorelu27 should also join the discussions once we have a pull-request.

juan-castrillon:
## First Version of the Master Class

This is a first attempt in the master class. There is still parts not very clear to me and therefore missing. This will eventually close #15 

 ### Important points for this commit:

- Master is implemented with a method that receives a `String` command ("deploy application") and returns a  `String` response ("Helm chart execution")

- I came up with 5 different commands to try, given the abstract level of the task.

- The commands are stored in a txt file inside `src/main/resources` in the following format: _command->response_

- To create the interface between the master and the commands, I followed the [class diagram](https://user-images.githubusercontent.com/12758306/89473539-6695c380-d783-11ea-965e-7ab919316ce9.png) in the following way:

  - Created a `MasterResource` interface: Done as a generic interface (different kind of resources) and with the abstract method `get()`

  - Created a `MasterConfig` interface: Also a generic interface (Configurations can be given in different types), appart from the `get()` method, added a `set()` method in order to be able to change configurations in runtime (not implemented yet)

  - Created a `CommandSet` class that implements the `MasterConfig` interface, and is implemented as a singleton to provide the master the commands

- Created a test file with tests for all the defined commands as well as empty string commands and not recognized commands


juan-castrillon:
@alyhasansakr Here is the corrected main class. 

I added the things we discussed such as: 

- Abstract Input and Output
- Concrete Implementation of those using console
- Configuration File as a JSON
- Main method (and loop) inside the master, that reads, maps and outputs constantly

For the JSON implementation, and based of some research I went with the [Jackson Library](https://github.com/FasterXML/jackson). Its fairly lightweight but very powerful and its implementation was very simple. However, as everything else with these class, is open to change if better options are found.

Which classes and interfaces are still needed or maybe could be deleted is not clear, but I believe its a good first approach. Like you said, this will be refactored a lot.

Issue #17, opened 2020-08-13
Scenarios


alyhasansakr:
A Scenario is a series of Events that are triggered in a timely manner. Scenarios are stored as files (JSON) which can be loaded and executed. Running a Scenario starts a thread that trigger Events, triggering an Event just feeds input to the master (which will be routed according to the configuration). Therefore, an Event should implement MasterInput.

As far as the description here goes, there is no need for specific Event classes (deploy app, stop app), as these should be just different inputs for the master, basically just text. The actual effect of an event would be a new MasterOutput for each event, we are going to write separate issues for each of them.

To edit a Scenario, we need a ScenarioEditor that implements MasterOutput. The ScenarioEditor should have the following functionalities:
1. Create a new Scenario.
1. Add an event to a Scenario.
1. Delete an event from a Scenario.
1. Save Scenario to file.
1. Read Scenario from file.

To execute a Scenario, we need a ScenarioDispatcher that implements MasterOutput. The ScenarioDispatcher should have the following functionalities:
1. Load a Scenario.
1. Run (or restart) a Scenario.
1. Pause a Scenario.

To close this ticket, a manual test procedure with console as input and output should be possible for both ScenarioEditor and ScenarioDispatcher.

PR #23, opened 2020-08-18
Scenarios v1

juan-castrillon:
## First attempt at scenarios

Here is where I am in solving #17. All the functionalities are still not complete, however, I wanted to start the pull request so you guys would be able to see my progress, and maybe give me some feedback, also determine whether I am in the right path or not.

### Right now, these are implemented:
- `Scenario `class based on a JSON File. 
- `Event `class implementing `MasterInput`
- `ScenarioEditor `class, completely functional, tested via the command line
- `ScenarioDispatcher` class, with only the `load scenario` functionality implemented

### Still missing: 
- ~~Event Triggering~~ (See last commit)
- Event Timing Characteristics (Right now, just an integer). Is it needed?
- Relate scenarios (events) with inputs in the master without making the configuration dynamic (this is the tricky one)
  (For this, in my last commit, I came up with a solution to make the scenario class a MasterInput implementation as well, and have it deliver each event everytime read is called)
- Run and Pause functionalities in `ScenarioDispatcher` (Also without changing the configuration of the master)
- ~~Test are failing right now, need to be redesigned based on the new functionalities~~

I will keep working on the missing parts, but any comment, suggestion or feedback is of course appreciated 😄


juan-castrillon:
After some other details, I think I may need some help for what is left. 

In this last commits, I turned the `Master` into a singleton, and used that to implement a run scenario capability in the scenario dispatcher. Still, in order to make everything right, I still run into the issue of needing runtime changes in the configuration of the master, which is not wanted. So, do any of you have any suggestions on how to approach this? . Also, the way its currently implemented, I cant think of a way to implement the pause scenario functionality. 

Thanks in advance for the help!

alyhasansakr:
There is many ways to do it, let me think about it, you don't need to do anything until then.

src/main/java/InfrastructureManager/ScenarioDispatcher.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
+
+    private void pauseScenario() {
+        //TODO: Implement
+    }
+
+    private void runScenario() {
+        Master master = Master.getInstance();
+        for (Event e : this.scenario.getEventList()) {
```
alyhasansakr, line 38: This loop goes into the master, here we call master.runScenario(this.scenario)

alyhasansakr: and it needs to be a separate class that is Runnable, then you would be able to run in a separate thread and pause it.

alyhasansakr: this means that the master main function should also do the same, just create and run several instances of this new class.

alyhasansakr: the configurator role in this case would be to predefine the outputs that would be connected to this input, for that you would need a scenario ID given in the configuration and the same ID used to run the scenario.

juan-castrillon: Hey @alyhasansakr  , here when you say :
> and it needs to be a separate class that is Runnable, then you would be able to run in a separate thread and pause it.
What exactly do you mean? What should be a separate runnable class? The master?

alyhasansakr: No, a new class, called MasterThread or something, it just contains this loop and runs in its own thread, and of course this needs to be generalized to the same loop in the original main.
In Scenario case, it would be created but not assigned an actual Scenario instance, until a Scenario is created, then the MasterThread instance should be started.

juan-castrillon:
Its still not finished, because the test are missing. But the functionality is correct according to what we have discussed.

src/main/java/InfrastructureManager/Master.java
```diff
@@ -1,37 +1,138 @@
...
+/**
+ * Master Class of the Infrastructure Manager, singleton class
+ */
 public class Master {
 
     private final CommandSet commandSet;
-    private final MasterInput input;
-    private final MasterOutput output;
+    private final ArrayList<Runner> runnerList;
+    private final ArrayList<Runner> running; //List to store currently running runners
```
alyhasansakr, line 12: I am not a fan of  using 2 lists, someone is going to start a thread and forget to add it to the other list.
There should be no problem in using 1 list, but if there is then use a small class for it.

src/main/java/InfrastructureManager/Master.java
```diff
@@ -1,37 +1,138 @@
...
+     */
+    public static Master getInstance() {
+        if (instance == null) {
+            instance = new Master();
+        }
+        return instance;
+    }
+
+    public static void main(String[] args) {
+        Master.getInstance().startMainRunner();
```
alyhasansakr, line 134: I think the main should wait for threads to finish (join).

src/main/java/InfrastructureManager/Master.java
```diff
@@ -1,37 +1,138 @@
...
+
+    /**
+     * Method that given a certain Scenario, runs the corresponding ScenarioRunner and adds
+     * it to the list of running Runners. If called on a running ScenarioRunner, it will restart
+     * it
+     * @param scenario Scenario to be run
+     */
+    public void runScenario(Scenario scenario) {
+        ScenarioRunner scenarioRunner = getRunner(scenario);
+        if (running.contains(scenarioRunner)) { //If running then stop it and re-run
```
alyhasansakr, line 67: why? if running just leave it running.

src/main/java/InfrastructureManager/Master.java
```diff
@@ -1,37 +1,138 @@
...
+    }
+
+    /**
+     * Method that given a certain Scenario, runs the corresponding ScenarioRunner and adds
+     * it to the list of running Runners. If called on a running ScenarioRunner, it will restart
+     * it
+     * @param scenario Scenario to be run
+     */
+    public void runScenario(Scenario scenario) {
+        ScenarioRunner scenarioRunner = getRunner(scenario);
```
alyhasansakr, line 66: Where is exception handling? (this is not the only case)

src/main/java/InfrastructureManager/Master.java
```diff
@@ -1,37 +1,138 @@
...
-        String mapping;
-        while (true) {
-            in = master.fromInput();
-            mapping = master.execute(in);
-            master.toOutput(mapping);
-            if (in.equals("exit")) {break;}
+    /**
+     * Method for starting the main runner thread, declared in the config file
+     */
+    //TODO:This could maybe go in configurator
```
alyhasansakr, line 48: I think it stays here.

src/main/java/InfrastructureManager/Master.java
```diff
@@ -1,37 +1,138 @@
...
+     */
+    private ScenarioRunner getRunner(Scenario scenario) {
+        ScenarioRunner scenarioRunner;
+        for (Runner runner : runnerList) {
+            try {
+                scenarioRunner = (ScenarioRunner) runner;
+                if (scenarioRunner.getScenarioName().equalsIgnoreCase(scenario.getName())){
+                    return scenarioRunner;
+                }
+            } catch (ClassCastException e) {
```
alyhasansakr, line 115: As a general rule, most of the time a general catch is better than specific ones.
Here we are missing the case where scenario is null, just change to catch (Exception e).

src/main/java/InfrastructureManager/Master.java
juan-castrillon: I was thinking about whats written in the issue #17 as a wanted feature for the ScenarioDispatcher:

> Run (or restart) a Scenario.

So, in this way, by calling run again on a scenario, it would restart.

src/main/java/InfrastructureManager/Master.java
juan-castrillon: You are right, I've already solved it by getting rid of the `runnning` list, and getting that info direclty from the Runner. I will resolve this comment

src/main/java/InfrastructureManager/Master.java
juan-castrillon: I thought this was the simplest way to achieve the join, but I am not 100% sure is the best. I didnt wanna expose or save the other threads, given that the exit command is made for closing all of them

juan-castrillon:
I still havent fixed the run and re run waiting for your input @alyhasansakr , but apart from that (And maybe also adding more tests), this first version seems ready

src/main/java/InfrastructureManager/Master.java
alyhasansakr: My concern is if we have two different inputs that try to start the same event, then the second one may unintentionally restart it, I think we need a stop function.
By the way, the design in the original issue is preliminary and may change on the way based on problems we face. The important part is that we don't "just fix the problem", instead we alter the design and make sure that the changes are reflected everywhere not just where the problem occurs.

alyhasansakr:
Testing is good for now, but I am still missing a timing solution for Events.
I think the best solution is to pass the start time of a Scenario as a parameter when running the Scenario (absolute time), then every Event has a time (defined in the Scenario file) relative to the Scenario start time.
Don't use delay as it accumulates timing error (coming from execution time), instead use a wait until function.
If a Scenario is given a start time in the past it should throw an exception.

src/main/java/InfrastructureManager/Master.java
juan-castrillon: You are right, I havent considered that case. I will add the stop function, also in the dispatcher, and allow the running runners to just keep running. 👍

juan-castrillon:
@alyhasansakr  Something is still not clear to me regarding the timing : If Events are sequential inside an scenario (happen one after the other and cannot happen in parallel) , isnt guaranteeing the execution order to be the same as in the Scenario file enough? Or am I missing something? In which case then would we need to define times?

alyhasansakr:
@juan-castrillon 
By timing I mean a certain point of time, for example, event 1 occurs after 1 second of scenario start, and event 2 occurs after 5 seconds of scenario start.
Scenario start is a set when the scenario is started, and it should be absolute time (number of milliseconds since 00:00:00 UTC on 1 January 1970).

juan-castrillon:
The only thing missing right now, is passing the start time as an argument while running. In the last commit, everything is made to pass the current absolute time to the scenario (`System.currentTimeMillis()`) and run it "right away". I didn't implement it yet, because I wanted some input in whether that was a good idea, to make the start time an argument. Wouldnt we want to start scenarios as soon as we call run_scenario on the master? I think for example in the console input mode, is not very practical to input for example 1444148155508 everytime we want to run. 

So my guess would be either, let it with the current time, or define a relative time in the scenario file as well (Same as with the events)

alyhasansakr:
I would say it should be a parameter and support both cases (relative and absolute). So I normally would send something like "run_scenario after 10 seconds" (or whatever format you wish to use), but also keep the absolute time capability for other types of inputs.

juan-castrillon:
I am running into a curious thing. Running locally everything works fine, testing passes as well in my local environment, but not in the github actions. I know it has something to do with the "run now" mode when running an scenario, where the comparison for the past values is made, the start time passed by the machine is concluded to be a past value and the exception is thrown, therefore the test `runScenarioTest()` is failing. I believe adding a "time window" between the two values (present and start time) will fix it, but not sure if that is the right solution. I try adding a 10 ms value and it still failed in github and 1 second and it passes. @alyhasansakr Can you think of anyway of doing this in a good way?

alyhasansakr:
This has to do with how fast the test system is (GitHub is slow), anyway I would have set like 2 seconds delay as the default (when no delay is specified), the way you do it is the right way.

juan-castrillon:
Ok, so I defined it as a constant instead, and I think then, its ready to be merged

alyhasansakr:
Do we need the files in "out"?

src/main/java/InfrastructureManager/ScenarioRunner.java
```diff
@@ -0,0 +1,78 @@
...
+        currentEvent = 0;
+        super.exit();
+    }
+
+    /**
+     * Wait for executing the states according to their relative execution times
+     * @param e Event to be waited for
+     */
+    private void waitForEvent (Event e) {
+        while (true) {
```
alyhasansakr, line 71: That is a busy waiting loop, the upside is that it gives highest precision possible, the downside is that it uses the CPU all of the time, which is not a good design choice when we expect a lot of these running at the same time.
You have the following solutions ordered by simplicity:
1. Add a little bit of delay each iteration (1 ms will do).
1. Calculate remaining time and sleep for exactly that time.
1. Same as above but sleep only half the period, repeating until the remaining time is short enough, then continue busy waiting (still with a short delay).

I prefer the second one, but you may do whichever you see fit.

juan-castrillon:
Ok done. Just as a question @alyhasansakr, do you think its worth considering changing the timing implementation to methods like `System.nanoTime()` or using the 'Instant' class, to improve precision or performance? I am not sure based on what I have read

alyhasansakr (approved):

alyhasansakr:
No need to use any other timing library for now, our required precision is about 10 ms, so should be good.

Issue #29, opened 2020-09-05
AdvantEdge Integration


alyhasansakr:
Implement a controller for AdvantEdge. It uses REST to send commands to AdvantEdge.

The main functionalities that we need are:
- Deploy an AdvantEdge scenario.
- Terminate an AdvantEdge scenario.

Tests may not include an actual AdvantEdge deployment, only check if the REST requests are sent.

juan-castrillon:
After doing some research, there is two ways of achieving this, as can be seen [here](https://github.com/InterDigitalInc/AdvantEDGE/wiki/API-Documentation#interfacing-with-api) : 

1.  Make the requests from software directly to AdvantEDGE HTTP endpoints. This would integrate easily into our program, and the APIs are ver well defined [here](https://github.com/InterDigitalInc/AdvantEDGE/wiki/api-sandbox-ctrl) and [here](https://github.com/InterDigitalInc/AdvantEDGE/wiki/api-platform-ctrl) for the functionalities we need right now (and probably the ones we would need eventually) . The only downside i find is that in the AdvantEDGE API documentation, it says 
> This method works but is not optimal

2. The recommended method by AdvantEDGE is to use a [swagger code generator](https://github.com/swagger-api/swagger-codegen) to generate a client package inside our project based only on the YAML files describing the API. Although, this looks promising, I've tried it and I think it has two downsides: First, it will add all the functionality for each YAML file, that means a lot of files and code we wouldnt use. Second, it apparently bases its JSON implementation on the GSON library, so in order to have only one JSON library, the current implementation with Jackson would have to be changed.

What do you think @alyhasansakr , is it worth the complexity of the generated java package api over the simple request making?

alyhasansakr:
The first solution is optimal for us, go with it.

PR #32, opened 2020-09-11
Advantedge integration v1

juan-castrillon:
This is the first attempt at a pull request to solve #29.  In the commits the following was made:

- An Object implementing `MasterOutput` was added to act as a Client for the AdvantEDGE API. All necesary, changes in order to add a new output were made. 

- Using the built in 'HttpClient' package, and having in mind how to [interact ](https://github.com/InterDigitalInc/AdvantEDGE/wiki/api-platform-ctrl) [with ](https://github.com/InterDigitalInc/AdvantEDGE/wiki/api-sandbox-ctrl)AdvantEDGE API, the client can now perform POST and DELETE requests. In this last commit, the request are made to postman-echo, just as a proof of concept, and the body, as well as the response is printed on the console. 

Still missing, and very important for the integration with AdvantEDGE, is the definition of a JSON file describing the scenario, according to the model in the AdvantEDGE documentation, which need to go into the body of the POST requests.  To see this model, [this file](https://raw.githubusercontent.com/InterDigitalInc/AdvantEDGE/master/go-apps/meep-platform-ctrl/api/swagger.yaml) need to be imported (from url) in [a swagger editor](https://editor.swagger.io/).

Also, testing is missing.


juan-castrillon:
@alyhasansakr  I have a question on how to proceed with the advantEdge scenario definition. Should I manually make a JSON file compliant with the model and just send it in the request? Or create a whole bunch of classes to create the file from objects in the program? What do you think?

alyhasansakr:
@juan-castrillon For simplicity, we go with a single JSON file.
You can actually create this file using the interface of AdvantEDGE, that should be easier than writing it manually.
Let me know if you have any problems using AdvantEDGE on a virtual machine.

juan-castrillon:
Hey @alyhasansakr, ok, so I will create a new virtual machine and install and run advantage.  I will let you know, if i have any issues with that. Thanks for the feedback!

juan-castrillon:
After struggling, I was able to run AdvantEDGE and create a scenario (based on the first part of [this one](https://github.com/InterDigitalInc/AdvantEDGE/wiki/first-scenario#construct-scenario-physical-infrastructure)). However, AdvantEDGE exports the scenarios in a YAML file, so I implemented in our program, a parser to convert this YAML files to JSON, and thus be able to send them in the body of the requests. In the generated YAML file, there is data missing, if compared with the model of a scenario defined in the API, but I believe, they are optional values. 

Testing, is the only thing missing, but because right now, the requests are send to the Postman Echo service, I was able to run the master and prove that the request as well as the parsing is working, so I just have to formally do it as a test.

juan-castrillon:
I've added testing, using [Wiremock](http://wiremock.org/) to create a mock server, and test the request that are made. This is working in my machine, however, it isnt in the Github Actions one. To be exact, I get the following on the actions log:

> com.github.tomakehurst.wiremock.common.FatalStartupException at WireMockServer.java:149
        Caused by: java.lang.RuntimeException at JettyHttpServer.java:192
            Caused by: java.net.SocketException at Net.java:-2

Which I believe it to be due to the github machine not having port 80 open. But I am not completely sure, thats the issue (because locally I didnt had to open any ports for it to run), so before I start messing around the machine, I wanted to know your take on this @alyhasansakr @theodorelu27. Once this is figured out, I believe the function is ready to be merged

alyhasansakr:
@juan-castrillon I believe your assumption is correct.
Port 80 is open by default for most operating systems (it is HTTP port), however, it makes sense to disable this port for GitHub workers for security reasons.
To stay clear of this problem, just change the port number to any number above 10000 and open this port in the GitHub virtual machine.

juan-castrillon:
Ok, then I will do that. I just made it on port 80, because its the one that [advantEdge](https://github.com/InterDigitalInc/AdvantEDGE/wiki/API-Documentation#apis-overview) uses for its controller API, but I will change it and then open the ports on the worker. 

We''ll have to remember to change it again when deployed with advantEdge

src/test/java/InfrastructureManager/AdvantEdge/AdvantEdgeClientTests.java
```diff
@@ -0,0 +1,78 @@
...
+        String scenarioPath = "src/test/resources/AdvantEdge/dummy-test.json";
+        client.out("advantEdge create " + scenarioName + " " + scenarioPath);
+        verify(postRequestedFor(urlEqualTo(path))
+                .withRequestBody(
+                        matchingJsonPath("$.name", equalTo(scenarioName))
+                ));
+    }
+
+    @Test // Checks if adding the scenario as a YAML file, sends a JSON in the request
+    public void YAML2JSONParserTest() {
```
theodorelu27, line 35: since this test is more about the functionality of sending request with a YAML file and parsing isn't really the responsibility of AdvantEdgeClient, I think it'd be more clear if we name it as such
```suggestion
    public void addYAMLScenarioRequestTest() {
```

theodorelu27 (commented):
@juan-castrillon great work on the code, I think there's no issue with the functionality.
I had some comments, but it's mostly just basic stuff to make the code hopefully more readable, so hope it helps 😅

@alyhasansakr I'd also like a second opinion from you since this is my first time doing a formal code review, so I hope I didn't have the wrong focus while working on it

juan-castrillon:
@theodorelu27 Thanks for the corrections! Although small, I agree with you in all of them. See the new commits, each one fixing one of your review comments.

theodorelu27 (dismissed):
@juan-castrillon glad I could help
the changes looks good, so I'd give it my approval

src/main/java/InfrastructureManager/AdvantEdge/AdvantEdgeClient.java
```diff
@@ -0,0 +1,144 @@
...
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpRequest.BodyPublishers;
+import java.net.http.HttpResponse;
+import java.nio.file.Paths;
+import java.time.Duration;
+
+public class AdvantEdgeClient implements MasterOutput {
+    private final String requestPath = "http://localhost:10500"; //The controller API is exposed on port 80 & 443 of the node where AdvantEDGE is deployed.
```
alyhasansakr, line 15: I believe port 10500 is for testing only, if this is the case, we need to set this variable in the configuration as well

src/test/java/InfrastructureManager/AdvantEdge/AdvantEdgeClientTests.java
```diff
@@ -0,0 +1,79 @@
...
+import static com.github.tomakehurst.wiremock.client.WireMock.*;
+import static com.github.tomakehurst.wiremock.core.WireMockConfiguration.options;
+
+
+public class AdvantEdgeClientTests {
+    private final AdvantEdgeClient client = new AdvantEdgeClient();
+    private final String scenarioName = "dummy-test";
+    private final int portNumber = 10500;
+
+    @Rule //Mock server on port 80
```
alyhasansakr, line 19: correct comment

juan-castrillon:
You are right @alyhasansakr . In my last commit I fixed the port configuration 

So now if in the configuration file something like this is given : 

>  "outputs" : ["advantEdge port10500"]

The output will be to port 10500, however, if no argument is passed:

>  "outputs" : ["advantEdge"]    

Port 80 will be set by default

alyhasansakr (approved):

Issue #30, opened 2020-09-05
Cgroups Integration


alyhasansakr:
To limit CPU usage for edge nodes, we use Cgroups in a BASH script.

The script should be able to receive the name of an edge node (node-1 for example) and apply a defined CPU limit (1.5 cores for example). The main idea is that any running instance of an application should have the edge node name as part of the command that was used to start it, and thus is searchable.

Here is an [example](https://github.com/EdgeLab-FHDO/Edge-Diagnostic-Platform/files/5177819/limit.txt).

Don't forget to add automated tests.

juan-castrillon:
Hey @alyhasansakr. So just to be clear and as i am new to BASH scripts, I want to have things clear from the beggining:

 - I understand that I should add another MasterOutput, who is capable of running a bash script, given a command from the master.  The bash script runs in a different machine than the master right? So and additional component should be developed to be deployed in the same machine as the bash script and listen for a command from the MasterOutput (via REST?).

 - The bash script to run, should accept the edge node as a parameter and use cgroups to limit its CPU. It can be based on the example you provide. 

 - The automated test, should be on the java developed components (Or the bashscript?)

Am I getting everything right?

PR #34, opened 2020-09-24
Cgroups integration v1

juan-castrillon:
First attempt at solving #30 

So far, this is the implementation of the Cgroups Functionality in the master. In the last commit, its working like this:

- The rest server, now handles GET request to the `/limit` URL

- The RestOutput (that given that it was statically connected to everything, I made as a singleton) can now handle commands of the type `limit_node {tag} {limit}`

 - When the request is made (will be from the bash script), the response is a json string containing all the nodes that will be limited like this:

image

After this, I will start working on the second part of the functionality, that means the BASH script.


juan-castrillon:
In my last commit, the only thing missing for the functional part is the Assignment of the process based on the PID to the cgroup. I dont quite get how does that work in the example, could you please help me clarify it @alyhasansakr?

The partially finished bash script can be found for now under `src/main/resources/BASH/limit.sh`. I am also still not sure where to save it-

Apart from that, I've added automated test for the limit node functionality.

alyhasansakr:
@juan-castrillon The target is to get the process ID, you can use the command ps in Linux to get that, then you need to use awk to filter the output of ps to contain only the processes that you want to apply the limit to (running instances of the application), that is not possible unless there is a sort of tag in the command that was used to start the instance (one of the parameters passed to the .exe of the application server instance).

I used the following to start servers:
`./server 172.17.0.1 3002 server-zone-1-poa-1 2500 1 poa-1 zone-1 1000 10`
This string will be displayed in the output of the ps command, so you can filter with "zone-1-poa-1" as a tag to limit the CPU usage of all server application instances that run on zone-1 poa-1 location (AdvantEDGE location).

You will need to do a similar thing here, but my recommendation is to make it as general as possible, so don't specify "zone" in your script, it should be enough that the script uses a given tag ($node_tag that you already have) to search the ps output for process ID.

Also, the script is part of the source code, move it to a new folder bash under src/main.

juan-castrillon:
@alyhasansakr thanks for the help. With that and some reading, I was able to grasp a good concept on how those commands work. In my last commits:

- I changed the directory of the bash script to `src/main/bash/NodeLimit`
 - I implemented the clasification of the processes based on SPID (Thread ID) and not PID (like they did in the example you provided)

I tested the script  by running a dummy script that took an argument in 4 tabs and then in a 5th calling the script and print the SPIDs to the console. 

I believe, it is ready to be merged

src/main/bash/NodeLimit/limit.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+#
+#Script for creating a cpu limited cgroup based on information retrieved as a REST resource
+
+no_quotes_regex='"\K\w+(?=")' ##Regular expression to match a word between quotes, ignoring the quotes
+
+curl -s http://localhost:4567/limit | grep -Po '"\w+":"\w+"' > limitFile ##Do the GET request and save the data to a file
```
alyhasansakr, line 7: That will only read the limits once, we may have dynamic events that affect the limits, in such case the script should read from the REST resource again.

I would say make it a part of a loop (new one of course).

Another thing, make read the URL as a parameter of the script.

juan-castrillon:
You were right with your comments @alyhasansakr. In the last commits, the bash scripts now:

- Receives the URL as a parameter, and only works if one is passed. Otherwise, notifies the user
 - Runs in an infinite loop but only process data when is new data (Each time it processes, saves and compares before processing 
   again)

src/main/bash/NodeLimit/limit.sh
```diff
@@ -3,33 +3,40 @@
...
+          # number of cores = cfs_quota_us / cfs_period_us (e.g. 0.5 core = 50000 / 100000, 2 cores = 200000 / 100000)
+          sudo cgcreate -g cpu:/"$node_tag" #Create a cpu limited group with the name of the node
+          sudo cgset -r cpu.cfs_period_us="$period_us" "$node_tag"
+          sudo cgset -r cpu.cfs_quota_us="$quota_us" "$node_tag"
+
+          #Assignment of the processes to the group based on the PID
+          #First command in the pipe (ps) outputs all running processes (-e), in full format(-f) and the threads (-T) with SPID
+          #AWK command, returns the SPIDs of the process which were called with the node_tag as argument
+          # Xargs turns the output of awk into arguments for cgclassify, which then assigns the processes to the limited group
+          ps -ef -T | awk -v tag="$node_tag" '!/awk/ && $0 ~ tag { print $3 }' | xargs -t sudo cgclassify -g cpu:/"$node_tag"
```
alyhasansakr, line 34: A small problem here, if the process to be limited is not yet created (or the application restarted) the limit will not be applied.

This line must be executed even if there were no changes in the limits, it must be excluded from the if condition.

src/main/bash/NodeLimit/limit.sh
```diff
@@ -3,33 +3,40 @@
...
+      curl -s "$1" | grep -Po '"\w+":"\w+"' > limitFile #Do the GET request and save the data to a file
+      if [[ "$read_data" != $(cat limitFile) ]] #Only process if there is new data
+      then
+        while IFS=: read -r node_tag time_values #For each line of the file
+        do
+          node_tag=$(grep -Po "$no_quotes_regex" <<< "$node_tag") #Get rid of quotes
+          time_values=$(grep -Po "$no_quotes_regex" <<< "$time_values") #Get rid of quotes
+          IFS=_ read -r quota_us period_us <<< "$time_values" #Split both time values based on "_" separator
+
+          echo "Read tag: $node_tag"
```
alyhasansakr, line 20: small change here, only print if -v is given as a parameter.

src/main/bash/NodeLimit/limit.sh
juan-castrillon: I thought about that but I wasnt sure because, is it right to be constantly adding (each loop of the infinite while) the same (most of the time) processes to the group ?

alyhasansakr: You should always add a delay to the loop, lets say 1 second.

juan-castrillon:
Ok, the corrections were added in the last commit. The script now:

- Receives two kind of option parameters : `-v` activates verbose mode and `-a` is to be followed by the URL to make the get request. They can be passed to the script in any order (`-a` always followed by the URL)

 - Is always checking for processes (not only when there is new data) but to start doing that, waits until there is at least one cgroup created (data was processed once)

alyhasansakr (approved):

Issue #25, opened 2020-08-21
Simple OpenCV Tracking Application


alyhasansakr:
A simple pedestrian tracking application (or anything that moves) that uses OpenCV functions (CPU).

The application is divided into two parts: client and server, each part is a Docker application.

The client application reads video input (from a file) and is capable of doing all of the processing without the server. The client should provide the option to connect to a server or to just run locally. If connected to a server, the client should encode video frames before sending it to the server, then wait to receive the output from the server (detected objects).

The server application waits for a client connection, once connected it receives video frames from the client, decode it, run image processing on it then sends the result back to the client (detected objects).

This is the core implementation of the application, we don't need any connection to the master or a logger. Options such as client mode (whether to run locally or send to the server) should just be application parameters for now.

Image processing function has to be implemented in a separate class that is used in both client and server. Other than that you are free to implement any number of classes.

Don't forget about the unit tests. Also a manual test should be possible using Docker.

The original application can be found [here](https://github.com/EdgeLab-FHDO/ResourceAllocationE2EL), but you don't need to follow it exactly, it is really not good for what we want to do here.

theodorelu27:
@alyhasansakr before starting to work with it, I want to confirm the requirements for this issue:
1. Create a Client application
- can read video input
- can switch between local and server for client modes
- can encode video frames
- can send the encoded frames to the server (POST Request)
- can connect to a server
2. Create a Server application
- can decode incoming video frames
- send result back to the client (POST Request)

other / common traits:
- For the scope of this issue, we will only connect client with a known server, connection should nevertheless be abstracted
- The Client and Server should communicate with REST
- Both should have unit tests
- Both should have manual to test in Docker (put it in Wiki?)
- We would try to first implement this in Java, but if it's not possible, implementation with C++ is preferred

things I want to confirm:
- Can both the client and server each did more than one processing job?
- If a processing has already started in local and a server is connected, should it finish the current processing or should it move the processing to the server right away?

For the estimates, this is my first time doing this kind of application so I'd give it until Wednesday to confirm if it's viable to do in Java and finish it on Saturday, is it a realistic or otherwise estimate or is too long?

alyhasansakr:
First of all, I like your approach of identifying the requirements.

Let me answer the question first:
- Both client and server should not have more than one application instance, that is one video input and one detections output, both of them however, may have more than one thread to handle this instance. In other words, we use microservice model instead of a standard multi-threaded server.
[...]

PR #35, opened 2020-09-29
Open CV Docker Image Development

theodorelu27:
Initial Pull Request for #25 

Included are:
- Dockerfile to build an Ubuntu client with Java8 and OpenCV 4.5
- Sample codes for marker creation and detection
note: sample codes were added in the ``samples` directory to exclude it during build

Missing:
- Dockerfile not yet working
- Connection between nodes and server


theodorelu27:
@alyhasansakr I've found out the problem for the Docker image build, it's due to my Dockerfile being saved with CRLF (Windows End of File) instead of LF (Linux End of File)

so the build is working as intended now, I'd focus on the client-server connection now

theodorelu27:
@alyhasansakr I've pushed my most recent version

What has been done:
- Dockerfile to build an Ubuntu 16, OpenJDK8, OpenCV4.5 (Ubuntu and OpenJDK version may be changed)
- CreateMarker and DetectMarker
- Client that can be configured when running the class to use server or run locally
- Image File transfer between Client and Server over TCP using BufferedImage and ImageIO
- Server that handles and send back the detected marker image

What's still missing:
- Thread execution
- Master communication thread
- External configuration
- Configuration for cmake to reduce build time
- Unit testing

I haven't finished refactoring the sample codes and as of currently, the cmake still builds everything and takes around 1 hour to finish

I've also added the Test Files in a zip file in the samples directory which includes the image used in the test, as well as a pre-built opencv450 for Linux to cut down build time

samples/OpenCVClient/CreateMarker.java
```diff
@@ -0,0 +1,28 @@
...
+import org.opencv.aruco.Dictionary.*;
+
+import java.io.File;
+import java.util.Arrays;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Vector;
+
+public class CreateMarker {
+   public static void main(String[] args) {
```
alyhasansakr, line 17: I am with you that we need another a separate application to make the markers, but the location is not ideal, it should not be a part of the client.

My idea is that we use this "MarkerGenerator" to make images that we later use with the OpenCVClient. So images should be copied to the docker image that runs the OpenCVClient, and MarkerGenerator should not a part of that docker image.

samples/OpenCVServer/CreateMarker.java
```diff
@@ -0,0 +1,28 @@
...
+import org.opencv.aruco.Dictionary.*;
+
+import java.io.File;
+import java.util.Arrays;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Vector;
+
+public class CreateMarker {
+   public static void main(String[] args) {
```
alyhasansakr, line 17: See my other comment in OpenCVClient/CreateMarker.java

samples/OpenCVServer/OpenCVServer.java
```diff
@@ -0,0 +1,52 @@
...
+        try {
+            serverSocket = new ServerSocket(port);
+            clientSocket = serverSocket.accept();
+            out = new DataOutputStream(clientSocket.getOutputStream());
+            in = new DataInputStream(clientSocket.getInputStream());
+            currentImage = ImageIO.read(ImageIO.createImageInputStream(clientSocket.getInputStream()));
+            ImageIO.write(currentImage, "png", new File(fileName));
+            DetectMarker.detect(fileName);
+            currentImage = ImageIO.read(new File(resultFileName));
+            ImageIO.write(currentImage, "png", clientSocket.getOutputStream());
```
alyhasansakr, line 31: There is no need to send back the whole image, you should only send detected markers locations (and sizes).

alyhasansakr:
As you can see in the comments above, we need a little bit of restructuring:
1. There should be a separate application that generates a marker image (or video) and writes it to a file.
1. The client should read this previously generated file and send it to the server.
1. The server should detect the locations (and sizes) of the markers and send them (not the image) to the client.
1. Finally the client uses this information to modify the image.
1. Add an option for the client to show the annotated image in real time.
1. Image processing part should be done in one class that is independent of the application (image write and read should not be in the same function as detection and generation), this class should be used in all of the other 3 components (marker generator, client and server).
Let me know if you have any other questions.

Also we don't need to implement the master connection just yet, however, other parts you mentioned would be welcome.

theodorelu27:
@alyhasansakr thank you for the feedbacks, I've made some progress with it but still have a long way to go
so what I've done so far was:
- Separate MarkerGenerator into a separate application, 
- Separate ImageProcessing
- Update Client to reflect the new changes

what hasn't been done:
- Send informations (detected corners and ids) instead of full image to the client) 
I expect this to be done tomorrow so test with the previous version can be done first 
My approach would be to convert both value as a JSON and stream it back to the client which decodes the received JSON and draw the detected marker from the rebuilt information
- Video processing
- Real-time annotation with options 
so real time in this context would be to update the currently displayed image/video once the detected marker is available?
- Direct execution after build from Dockerfile (currently the CMD is disabled)
- Thread execution
- Sending files from generator to client 
so in this case the ImageGenerator wouild be a client and OpenCVClient would be a server
what I had in mind is the OpenCVClient would open a thread to only receive files and queue it
once it becomes available, the queue would be popped for processing
is this interpretation correct?
- External configuration
- Unit testing
- Configuration for cmake to reduce build time

a few notes:
- As we have now multiple classes using the OpenCV shared library which need to be called once, I currently called this in the main before anything else as a static method from DetectMarker, I'm note sure if by doing this I'm giving DetectMarker too many responsibility so alternative is calling the method as is in the main or create a class for it
- I haven't retested the Dockerfiles

I've tried to list the remaining tasks in the priority that I think is important but I'd need a feedback on the priority or if my interpretation is too complicated, specific questions about the tasks have been included in the corresponding tasks

alyhasansakr:
@theodorelu27 That is great, go on.
I have only one comment about "Sending files from generator to client", we don't need to do that in real-time, it is enough if the generator is ran one time only to create the files (images or video) then the files are copied to the docker image (using a dockerfile). Finally the client should just read the image files (or video file) normally, and just repeats doing so in an infinite loop.

theodorelu27:
@alyhasansakr thanks for the feedback, so here's the current to-do list:
- Send informations
- Video processing
- Real-time annotation with options
- External configuration
- Unit testing
- Configuration for cmake to reduce build time

theodorelu27:
@alyhasansakr I've updated the Dockerfiles and codes to prioritize testing with AdvantEdge
for now Parameter can be modified by updating the client_exe.sh and server_exe.sh
Image can be built for each by moving the Dockerfile in each directory to the parent manually or using batch script

There was a problem with the OpenCV building, I've tried running it with several configurations but it's not generating the .jar file that was needed to run the application, so for now the precompiled .jar and library is provided in Resource
Yes, this is a to-do list for me too to test the configurations, currently it's taking too long to compile the OpenCV jar

aside from that, there are no major changes to the code from our last test yet

src/main/java/Application/OpenCVClient/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+COPY build_classes.sh build_classes.sh
+COPY OpenCVClient/*.java /
+COPY Utilities/*.java /
+WORKDIR /
+RUN chmod +x build_classes.sh && sync && ./build_classes.sh && rm build_classes.sh
+
+#execute
+COPY OpenCVClient/client_exe.sh client_exe.sh
+WORKDIR /
+RUN chmod +x client_exe.sh && sync && ./client_exe.sh && rm client_exe.sh
```
alyhasansakr, line 21: In AdvantEDGE, the last line will be commented and written in another interface, this line must have the required parameters, what you are doing here is writing the parameters inside a file (client_exe.sh), which means that we cannot modify them just by modifying this line.

What you can do is to expose these parameters as parameters for client_exe.sh, so this line would be:
`RUN chmod +x client_exe.sh && sync && ./client_exe.sh && rm client_exe.sh CLIENT_ID=client13 MASTER_URL=http://host.docker.internal:4567/ BEAT_COMMAND=client/register GET_SERVER_COMMAND=client/get_node/`
This way we can use it with AdvantEDGE.

And do the same with the server.

theodorelu27:
@alyhasansakr in the recent commit I've exposed the parameter in the execution script, it can be modified by:
- setting environment variable
- uses provided flags when running the script

theodorelu27:
@alyhasansakr I've tried to do some fixes from our last discussion:
- updated the Dockerfile to accomodate the packaging
- moved reponsibility of Exception Handling from the utlity classes to larger classes
- change Runner extension to Runnable implementation
- separate Main and Method classes

I'm still not used to Exception handling, so I might went overboard and if there's anything else that I missed, please do point it out

src/main/java/Application/OpenCVServer/OpenCVServer.java
```diff
@@ -0,0 +1,47 @@
...
+        try {
+            activeServer.serverThread.join();
+            activeServer.heartBeatThread.join();
+        } catch (IllegalArgumentException e) {
+            e.printStackTrace();
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
```
juan-castrillon, line 45: Here again, if all handling is the same, just either do the more general (`Exception`) or the specific ones in one statement (with OR)

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
```diff
@@ -0,0 +1,164 @@
...
+    private String fileName;
+
+    private BufferedImage currentImage;
+
+    public ServerRunner serverRunner;
+    public HeartBeatRunner beatRunner;
+
+    private static OpenCVServerOperator instance = null;
+
+    public OpenCVServerOperator() {
```
juan-castrillon, line 40: for a singleton this should be private

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    public static OpenCVServerOperator getInstance() {
+        if (instance == null) {
+            instance = new OpenCVServerOperator();
+        }
+        return instance;
+    }
+
+    public void setupServerRunners(String args[]) throws IllegalArgumentException {
```
juan-castrillon, line 91: Again declaration in C style
```suggestion
    public void setupServerRunners(String[] args) throws IllegalArgumentException {
```

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
```diff
@@ -0,0 +1,164 @@
...
+            missingParameterList = missingParameterList + "BEAT_COMMAND,";
+        }
+        if(port == 0) {
+            missingParameter++;
+            missingParameterList = missingParameterList + "PORT";
+        }
+        if(missingParameter > 0) {
+            throw new IllegalArgumentException("Missing Parameter: " + missingParameterList);
+        }
+
```
juan-castrillon, line 154: Same comment as in the client operator. Consider changing this

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
```diff
@@ -0,0 +1,164 @@
...
+        String beatUrl = masterUrl + beatCommand;
+        String beatBody =  "{" +
+                "    \"id\": \"" + serverId + "\"," +
+                "    \"ipAddress\": \"" + serverIp + "\"," +
+                "    \"connected\": " + connected + "," +
+                "    \"port\": " + port +
+                "}";
```
juan-castrillon, line 161: Same comment as in the client operator, I believe the body construction is an internal concern of the beat

src/main/java/Application/OpenCVClient/OpenCVClientOperator.java
```diff
@@ -0,0 +1,269 @@
...
+        masterCommunicationRunner = new MasterCommunicationRunner(masterCommunicationUrl);
+    }
+
+    public void detectMarkerInServer() throws JsonProcessingException, InvalidObjectException, IOException {
+        System.out.println("detectMarkerInServer");
+        // connect to server and detect marker
+        String response;
+        JsonNode node;
+
+        startConnection();
```
juan-castrillon, line 220: Calling this in a loop (`ProcessingRunner`) creates a new socket everytime. I dont think thats very efficient given that data like IP and port dont change, so you will be creating the same socket every iteration

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    // change name to something more appropriate
+    public void standbyForConnection() throws InvalidObjectException, IllegalArgumentException, JsonProcessingException, IOException {
+        //if input is faulty, timeout and continue
+        currentImage = ImageIO.read(ImageIO.createImageInputStream(clientSocket.getInputStream()));
+        ImageIO.write(currentImage, "png", new File(fileName));
+        Mat subject = ImageProcessor.getImageMat(fileName);
+
+        detector = new DetectMarker();
```
juan-castrillon, line 58: This is called in a loop, so you are creating  a new detector every time. Why not just create it once and then use it all the other times? (In the constructor for example)

src/main/java/Application/OpenCVClient/OpenCVClientOperator.java
```diff
@@ -0,0 +1,269 @@
...
+
+    private void handleServerDetectException(Exception e) {
+        e.printStackTrace();
+        detectMarkerInClient();
+    }
+
+    public void markerDetection() {
+        Mat result;
+        subject = ImageProcessor.getImageMat(fileName);
+        detector = new DetectMarker();
```
juan-castrillon, line 246: Again, called in a loop,  so a new detector is called every time

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
```diff
@@ -0,0 +1,164 @@
...
+
+import Application.Utilities.DetectMarker;
+import Application.Utilities.HeartBeatRunner;
+import Application.Utilities.ImageProcessor;
+import Application.Utilities.OpenCVUtil;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import org.opencv.core.Mat;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
```
juan-castrillon, line 21: can be optimized

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
```diff
@@ -0,0 +1,164 @@
...
+    public OpenCVServerOperator() {
+        fileName= "read.png";
+    }
+
+    public void startConnection() throws IOException {
+        serverSocket = new ServerSocket(port);
+        clientSocket = serverSocket.accept();
+        out = new PrintWriter(clientSocket.getOutputStream(), true);
+        in = new DataInputStream(clientSocket.getInputStream());
+    }
```
juan-castrillon, line 49: Maybe separate this behavior? Because this is also called in a loop, and in this loop:
- You want `.accept()` to be called every time because this would block until a connection is made
- You **DONT** want to create the server socket every time in the same port (just create it once)

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
```diff
@@ -0,0 +1,164 @@
...
+
+        out.println(result);
+
+    }
+
+    public void stopConnection() throws IOException {
+        in.close();
+        out.close();
+        clientSocket.close();
+        serverSocket.close();
```
juan-castrillon, line 81: Again, do you wanna close the server socket every iteration? Or just at the end?

src/main/java/Application/OpenCVServer/ServerRunner.java
```diff
@@ -0,0 +1,35 @@
...
+            } catch (IllegalArgumentException e) {
+                e.printStackTrace();
+            }  catch (InvalidObjectException e) {
+                e.printStackTrace();
+            }  catch (JsonProcessingException e) {
+                e.printStackTrace();
+            }  catch (IOException e) {
+                e.printStackTrace();
+            } catch (Exception e) {
+                e.printStackTrace();
```
juan-castrillon, line 31: Check the other comments about exceptions.

src/main/java/Application/OpenCVServer/ServerRunner.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class ServerRunner implements Runnable {
+    private OpenCVServerOperator activeOperator;
+
+    public ServerRunner() {
+        activeOperator = OpenCVServerOperator.getInstance();
+    }
+    @Override
+    public void run() {
+        while(true) {
```
juan-castrillon, line 17: Check out the comments on the ServerOperator to maybe modify this loop (dont do everything inside). Also, have in mind this loop can be exited so the thread woul never end (join) unless an exception happens or is killed

src/main/java/Application/OpenCVServer/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+COPY build_classes.sh build_classes.sh
+COPY OpenCVServer/*.java /Application/OpenCVServer/
+COPY Utilities/*.java /Application/Utilities/
+WORKDIR /
+RUN chmod +x build_classes.sh && sync && ./build_classes.sh && rm build_classes.sh
+
+#execute
+COPY OpenCVServer/server_exe.sh server_exe.sh
+WORKDIR /
+RUN chmod +x server_exe.sh && sync && ./server_exe.sh && rm server_exe.sh
```
juan-castrillon, line 20: Parameters are not passed for `server.sh`

juan-castrillon (commented):
Hi @theodorelu27 ! Check out my comments and let me know if you agree or if you think I made a mistake somewhere. As final comments:

- Consider encapsulation and data-hiding based practices and check out what can be private (fields and also methods)
- Consider making some of the variables `final`
- Check out the warnings in IntelliJ. Import Optimization for example can be done automatically with one click.
- A good idea once you are done with exception handling and changing the code maybe to run a dependency analysis on IntelliJ. That you know what exactly you need for the docker containers to be as lightweight as possible

Let me know if you need any help!

settings.gradle
```diff
@@ -1,2 +1,3 @@
 rootProject.name = 'Edge-Diagnostic-Platform'
 
+include ':libraries:opencv'
```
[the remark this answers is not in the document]
theodorelu27: it's from when I tried to add the OpenCV LIbrary, I've removed it and it caused no problem, thanks for pointing out!

src/main/java/Application/OpenCVClient/OpenCVClientOperator.java
```diff
@@ -0,0 +1,269 @@
...
+            ipLock.release();
+        }
+    }
+
+    public String sendImage() throws IOException {
+        String resp = "";
+        Mat ids;
+        currentImage = ImageIO.read(new File(fileName));
+        ImageIO.write(currentImage, fileExtension, clientSocket.getOutputStream());
+        System.out.println("sendImage3");
```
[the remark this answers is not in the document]
theodorelu27: so for this part, do you suggest to remove all or only part of the code?

src/main/java/Application/OpenCVClient/OpenCVClientOperator.java
```diff
@@ -0,0 +1,269 @@
...
+        resp = in.readLine();
+
+        return resp;
+    }
+
+    public void stopConnection() throws IOException {
```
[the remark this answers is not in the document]
theodorelu27: for this, I will add the call to stopConnection after the response has been recorded

src/main/java/Application/OpenCVClient/OpenCVClientOperator.java
```diff
@@ -0,0 +1,269 @@
...
+    private String resultFileName = "detected.png";
+    private String fileExtension = "png";
+    private Mat subject;
+    private BufferedImage currentImage;
+
+    public Semaphore ipLock;
+    public Semaphore serverLock;
+
+    //Server Communication Properties
+    private int utilizeServer;
```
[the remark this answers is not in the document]
theodorelu27: it was initially a parameter received as the main argument, so I used int for more convenience
but since we've updated other input methods where this convenience is no longer relevant, I agree with you that boolean could work better

src/main/java/Application/OpenCVClient/OpenCVClientOperator.java
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    public void setServerUtilization(int input) {
+        try {
+            serverLock.acquire();
+            utilizeServer = input;
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        } catch (Exception e) {
+            e.printStackTrace();
```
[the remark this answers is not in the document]
theodorelu27: I see, I always added the general Exception in the end in case there's anything that I might have missed, but since there is only one specific, I agree that it's more confusing to add a general exception here

as for the setServerUtilization to false, I don't think it's wise since the Exception happened when acquiring the lock, so if we modify the value without ensuring that we were able to secure the lock to do so, I think it kinda defeats the purpose
that being said I'm not too familiar with distributed systems so I would like to hear your opinion on this as well

src/main/java/Application/OpenCVServer/Dockerfile
theodorelu27: this was intended because the parameters would be set as Environmental Variable in AdvantEdge and will be loaded from there

src/main/java/Application/Utilities/DetectMarker.java
```diff
@@ -0,0 +1,54 @@
...
+
+   public Mat getIds() {
+      return ids;
+   }
+
+   public void setIds(Mat ids) {
+      this.ids = ids;
+   }
+
+   public static void initOpenCVSharedLibrary() {
```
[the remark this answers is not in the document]
theodorelu27: more because it's also used in the Server
that being said, it might be better to move it to OpenCVUtil, what do you think?

src/main/java/Application/Utilities/OpenCVUtil.java
```diff
@@ -0,0 +1,138 @@
...
+        ArrayNode dataNode = mapper.createArrayNode();
+        String returnedResult = "";
+
+        result.put("type", type);
+        result.put("rows", rows);
+        result.put("cols", cols);
+
+        if(type == CvType.CV_32S || type == CvType.CV_32SC2 || type == CvType.CV_32SC3 || type == CvType.CV_16S) {
+            int[] data = new int[cols * rows * elemSize];
+            subject.get(0, 0, data);
```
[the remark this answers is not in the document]
theodorelu27: it is from OpenCV Java, I'm not really sure how it works since the documentation doesn't say anything except for the parameter, but it seems the get is used to set the data field for Mat

multiple data types were used because each OpenCV type uses specific data types so I added everything to accomodate if we want to add new functionalities in the future

src/main/java/Application/testing_manual.md
```diff
@@ -0,0 +1,22 @@
...
+Run the client
+<br>
+``docker run --network [network_name] -it [client_image]``
+
+<b>Reruning network on existing containers]</b>
+<br>
+``Connect the server to network - docker connect --alias [server_alias] [server_image]``<br>
+``Connect the client to network - docker connect [client_image]``
+
+Default ``[server_alias]`` is "server", you may change this in OpenCVClient.java for testing
```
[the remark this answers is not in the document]
theodorelu27: this is an old manual, I'd need to rework or remove it since now we're no longer using  custom network for testing

src/main/java/Application/OpenCVClient/OpenCVClientOperator.java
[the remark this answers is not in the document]
theodorelu27: I just read you other comments, I'd address this when fixing the threads and connections

src/main/resources/Configuration.json
```diff
@@ -106,6 +106,13 @@
         "update_gui" : "console GUIUpdateExecution" ,
         "node_request" : "console matchMakingExecution"
       }
+    },
+    {
+      "in" : "console_in",
+      "out" : "match_m",
+      "commands" : {
+        "assign_client $client_id" : "matchMaker assign_client $client_id"
+      }
```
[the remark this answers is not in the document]
theodorelu27: I've pulled the Master and there seemed to be no new changes, is there and other indicator that you can suggest to confirm that the Configuration in my current branch is up to date?

theodorelu27:
@juan-castrillon thank you for taking the time to review my code, I agree with most of your suggestions and I've left comments on where I need your feedback on the suggestions

resolved conversations means I've included them in the recent commit and for the rest (the Exception and Threading) I'd address it later today

src/main/java/Application/OpenCVClient/OpenCVClientOperator.java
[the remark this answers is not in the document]
juan-castrillon: @theodorelu27 In this comment, I was talking about the print out. Do you need it for anything?

src/main/java/Application/OpenCVClient/OpenCVClientOperator.java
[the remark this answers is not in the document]
juan-castrillon: You are right about modifying the value, it beats the purpose of the lock.

Now that you mention, I wanted to ask you about the locks. If the only thing that uses this resources is a singleton, shouldn't the lock be on the singleton instance itself, not inside? 

The locks are added to ensure that only one thread can access the resource (in this case `utilizeRunner`) at the same time, and that way avoid two threads working with different values. But if the resource is modified by a singleton, by definition, there would be only one instance accessing the resources (independent from the number of threads). However, the singleton itself becomes the shared resource. So I think it would make more sense to protect it than the resources. Do you see my point? Do you agree?

src/main/java/Application/Utilities/DetectMarker.java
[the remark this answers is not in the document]
juan-castrillon: I agree, that class already only has static methods and it would make more sense to have it there

src/main/resources/Configuration.json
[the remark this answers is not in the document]
juan-castrillon: For example, the match maker declaration now looks like:
```json
{"type": "MatchMaker-Random","name": "match_m"}
```

src/main/java/Application/OpenCVClient/OpenCVClientOperator.java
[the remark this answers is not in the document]
theodorelu27: ah, nope it was for a previous testing and it has been removed

src/main/java/Application/OpenCVClient/MasterCommunicationRunner.java
```diff
@@ -0,0 +1,52 @@
...
+                e.printStackTrace();
+            } catch (IllegalArgumentException e) {
+                e.printStackTrace();
+            } catch (ArrayIndexOutOfBoundsException e) {
+                e.printStackTrace();
+            } catch (IOException e) {
+                e.printStackTrace();
+            } catch (Exception e) {
+                e.printStackTrace();
+            }
```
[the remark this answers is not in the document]
theodorelu27: yes for now we only print StackTrace
evaluation is still an abstract class, but if from the beginning the connection is not good, then it'd set ServerUtilization to false then contiue the loop to retry getting the server info

src/main/resources/Configuration.json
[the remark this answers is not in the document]
theodorelu27: okay, it should be up to date now

src/main/java/InfrastructureManager/EdgeNode.java
```diff
@@ -4,17 +4,20 @@
     private String id;
     private String ipAddress;
     private boolean connected;
+    private int port;
```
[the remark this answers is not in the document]
theodorelu27: yeah, I agree if it's used as ip:port in all other places then it might be more sensible to let them be and separate them in the application as it is a specific requirement within the application

theodorelu27:
@juan-castrillon thanks for the review, I think I've addressed all changes that I think is required
as for the infinite loop, it is due to Docker applications nature so for now we'd leave it be, but I've added the exit variables as a reference for future members if they're adapting it to non Docker applications
regarding the shared lock, we want to enable other threads to modify other shared resources, so we'd leave it be for now

@alyhasansakr I think you can start the re-review

src/main/java/Application/OpenCVClient/MasterCommunicationRunner.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public MasterCommunicationRunner(String masterUrl) {
+        activeOperator = OpenCVClientOperator.getInstance();
+        evaluation = new ConnectionEvaluator();
+        communicator = new MasterCommunicator(masterUrl);
+    }
+
+    @Override
+    public void run() {
+        while(true) {
```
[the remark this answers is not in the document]
alyhasansakr: This is a Docker application, so no need for this.
But anyways just add a TODO here

src/main/java/Application/OpenCVClient/ProcessingRunner.java
```diff
@@ -0,0 +1,15 @@
...
+
+public class ProcessingRunner implements Runnable {
+    private OpenCVClientOperator activeOperator;
+
+    public ProcessingRunner() {
+        activeOperator = OpenCVClientOperator.getInstance();
+    }
+    @Override
+    public void run() {
+        while(true) {
```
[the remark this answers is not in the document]
alyhasansakr: This is a Docker application, so no need for this.
But anyways just add a TODO here

src/main/java/Application/OpenCVClient/ProcessingRunner.java
```diff
@@ -0,0 +1,19 @@
+package Application.OpenCVClient;
+
+public class ProcessingRunner implements Runnable {
+    private final OpenCVClientOperator activeOperator;
+
+    //stub for future classes
```
alyhasansakr, line 6: Add TODO:

src/main/java/Application/OpenCVClient/OpenCVClientOperator.java
```diff
@@ -0,0 +1,219 @@
...
+import java.util.concurrent.Semaphore;
+
+public class OpenCVClientOperator {
+    private Socket clientSocket;
+    private BufferedReader in;
+    private DataOutputStream out;
+    private DetectMarker detector;
+    private ObjectMapper mapper;
+
+    private final String fileName= "singlemarkerssource.png";
```
alyhasansakr, line 23: Why fixed values? at least add a TODO

src/main/java/Application/OpenCVClient/OpenCVClientOperator.java
```diff
@@ -0,0 +1,219 @@
...
+            e.printStackTrace();
+        } finally {
+            serverLock.release();
+        }
+
+    }
+
+    public void setupTcpConnection(String ipAddress) throws IllegalArgumentException, InterruptedException {
+        String[] argument;
+        // make a rest request to the master to ask for server and port for server
```
alyhasansakr, line 107: If this is a todo write TODO:

src/main/java/Application/OpenCVClient/OpenCVClientOperator.java
```diff
@@ -0,0 +1,219 @@
...
+                throw new IllegalArgumentException("Invalid argument");
+            }
+        } finally {
+            ipLock.release();
+        }
+    }
+
+    public void setupClientRunners(String[] args) throws IllegalArgumentException {
+        String[] argument;
+        String clientId = ""; // for test: client13
```
alyhasansakr, line 125: Not a good place to write an example.

src/main/java/Application/OpenCVClient/OpenCVClientOperator.java
```diff
@@ -0,0 +1,219 @@
...
+        }
+        String beatUrl = masterUrl + beatCommand;
+        String beatBody =  "{\"id\" : \"" + clientId + "\"}";
+        beatRunner = new HeartBeatRunner(beatUrl, beatBody);
+        String masterCommunicationUrl = masterUrl + getServerCommand + clientId;
+        masterCommunicationRunner = new MasterCommunicationRunner(masterCommunicationUrl);
+    }
+
+    public void detectMarkerInServer() throws RemoteExecutionException, IOException {
+        System.out.println("detectMarkerInServer");
```
alyhasansakr, line 170: Not a good idea to write to console in a multi-threaded application.

src/main/java/Application/OpenCVClient/OpenCVClientOperator.java
```diff
@@ -0,0 +1,219 @@
...
+        try {
+            if(!connected) {
+                startConnection();
+            }
+            response = sendImage().replace("\\\\", "");
+
+            node = mapper.readTree(response);
+            Mat ids = OpenCVUtil.deserializeMat(node.get("ids").asText());
+            List<Mat> corners = OpenCVUtil.deserializeMatList(node.get("corners").asText());
+            System.out.println("Marker Drawn with remote informations");
```
alyhasansakr, line 183: Not a good idea to write to console in a multi-threaded application.

src/main/java/Application/OpenCVClient/OpenCVClientOperator.java
```diff
@@ -0,0 +1,219 @@
...
+            detector.setCorners(corners);
+        } catch (IOException | InterruptedException e) {
+            stopConnection();
+            throw new RemoteExecutionException(e);
+        }
+    }
+
+    public void detectMarkerInClient() {
+        detector.detect(subject);
+        System.out.println("Marker Drawn locally");
```
alyhasansakr, line 195: Not a good idea to write to console in a multi-threaded application.

src/main/java/Application/OpenCVServer/ServerRunner.java
alyhasansakr: This is a Docker application, so no need for this.
But anyways just add a TODO here

src/main/java/Application/OpenCVServer/ServerRunner.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+
+import java.io.IOException;
+import java.io.InvalidObjectException;
+
+public class ServerRunner implements Runnable {
+    private final OpenCVServerOperator activeOperator;
+
+    //stub for future classes
```
alyhasansakr, line 12: Add TODO

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
```diff
@@ -0,0 +1,141 @@
...
+
+    private BufferedImage currentImage;
+
+    public ServerRunner serverRunner;
+    public HeartBeatRunner beatRunner;
+
+    private static OpenCVServerOperator instance = null;
+
+    private OpenCVServerOperator() {
+        fileName= "read.png";
```
alyhasansakr, line 39: Why a fixed value?

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    public void startConnection() throws IOException {
+        serverSocket = new ServerSocket(port);
+        clientSocket = serverSocket.accept();
+        out = new PrintWriter(clientSocket.getOutputStream(), true);
+        in = new DataInputStream(clientSocket.getInputStream());
+    }
+
+    // change name to something more appropriate
```
alyhasansakr, line 50: Add TODO:, or just change it to processing.

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
```diff
@@ -0,0 +1,141 @@
...
+    public static OpenCVServerOperator getInstance() {
+        if (instance == null) {
+            instance = new OpenCVServerOperator();
+        }
+        return instance;
+    }
+
+    public void setupServerRunners(String[] args) throws IllegalArgumentException {
+        String[] argument;
+        String masterUrl = ""; // for test: http://host.docker.internal:4567/
```
alyhasansakr, line 90: Not a good place to write examples.

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
```diff
@@ -0,0 +1,141 @@
...
+                        break;
+                    case "MASTER_URL":
+                        masterUrl = argument[1];
+                        missingParameterList.remove("MASTER_URL");
+                        break;
+                    case "BEAT_COMMAND":
+                        beatCommand = argument[1];
+                        missingParameterList.remove("BEAT_COMMAND");
+                        break;
+                    case "CONNECTED":
```
alyhasansakr, line 120: Why is that a parameter?

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
```diff
@@ -0,0 +1,141 @@
...
+        detector.detect(subject);
+        String ids = OpenCVUtil.serializeMat(detector.getIds());
+        String corners = OpenCVUtil.serializeMatList(detector.getCorners());
+        String result = "";
+
+        ObjectMapper mapper = new ObjectMapper();
+        ObjectNode resultObject = mapper.createObjectNode();
+        resultObject.put("ids", ids);
+        resultObject.put("corners", corners);
+        System.out.println("responding");
```
alyhasansakr, line 66: Printing in a multi-threaded application is not a good idea.

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
```diff
@@ -0,0 +1,141 @@
...
+    public void startConnection() throws IOException {
+        serverSocket = new ServerSocket(port);
+        clientSocket = serverSocket.accept();
+        out = new PrintWriter(clientSocket.getOutputStream(), true);
+        in = new DataInputStream(clientSocket.getInputStream());
+    }
+
+    // change name to something more appropriate
+    public void standbyForConnection() throws IllegalArgumentException, IOException {
+        //if input is faulty, timeout and continue
```
alyhasansakr, line 52: I don't understand that.

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
```diff
@@ -0,0 +1,141 @@
...
+        currentImage = ImageIO.read(ImageIO.createImageInputStream(clientSocket.getInputStream()));
+        ImageIO.write(currentImage, "png", new File(fileName));
```
alyhasansakr, line 54: Give me a good reason to write to a file instead of just running the detector directly (after conversion of course).

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
alyhasansakr: Of course not every iteration, that would not be good for latency.

src/main/java/Application/OpenCVServer/ServerRunner.java
```diff
@@ -0,0 +1,31 @@
...
+    private volatile boolean running = false; //Flag to see runner status
+
+    public ServerRunner() {
+        activeOperator = OpenCVServerOperator.getInstance();
+    }
+    @Override
+    public void run() {
+        while(true) {
+            try {
+                activeOperator.startConnection();
```
alyhasansakr, line 23: Do the same as you did with the client, only start a connection if not connected and close only on errors.

src/main/java/Application/Utilities/HeartBeatRunner.java
```diff
@@ -0,0 +1,34 @@
+package Application.Utilities;
+
+import java.io.IOException;
+
+public class HeartBeatRunner implements Runnable {
+    private final EdpHeartbeat beater;
+
+    //stub for future classes
```
alyhasansakr, line 8: Add TODO

src/main/java/Application/Utilities/HeartBeatRunner.java
```diff
@@ -0,0 +1,34 @@
...
+    private volatile boolean exit = false; //Flag to check status and be able to exit
+    private volatile boolean running = false; //Flag to see runner status
+
+    public HeartBeatRunner(String url, String body) {
+        beater = new EdpHeartbeat(url, body);
+    }
+    @Override
+    public void run() {
+        while(true) {
```
alyhasansakr, line 17: Add TODO for use of exit and running.

src/main/java/Application/Utilities/EdpHeartbeat.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.io.IOException;
+import java.net.ConnectException;
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.time.Duration;
+
+public class EdpHeartbeat {
```
alyhasansakr, line 11: What is "Edp"?

src/main/java/Application/Utilities/DetectMarker.java
```diff
@@ -0,0 +1,46 @@
...
+   private List<Mat> corners;
+   private Mat ids;
+
+   public DetectMarker() {
+      this.corners = new ArrayList<>();
+      this.ids = new Mat();
+   }
+
+   public void detect(Mat subject) {
+      Dictionary dictionary = Aruco.getPredefinedDictionary(Aruco.DICT_6X6_250);
```
alyhasansakr, line 20: Why fixed value?

alyhasansakr (changes requested):
Other general comments:

Application files should be under src/applications/markerdetection/ and properly separated in folders java, resources, bash, docker. Also, remove files that are no longer needed (testfiles.7z for example). Make sure that everything is still working after you move the files.

As you can see there are a lot of changes that I just ask you to add a TODO instead of implementing it now. You can also do that even for the ones that I didn't specifically ask for a TODO for (unless it is essential).

For writing to console, add a parameter "debug_mode" which enables it.

Your extensive use of global variables is not a good practice, you should use more function parameters. For example, detectMarkerInClient and detectMarkerInServer should receive a Mat parameter rather than using an instance variable. Another example is making OpenCVClientOperator a Singleton for no good reason. No need to change anything here, but keep that in mind for future implementations.

src/main/java/Application/Utilities/EdpHeartbeat.java
theodorelu27: EDP is Edge Diagnostic Platform, it's a placeholder name so can be changed if needed

src/main/java/Application/Utilities/DetectMarker.java
theodorelu27: up until this point I only tested with that particular dictionary
but there are not particular reason so it can be taken out as a class variable or loaded from configuration (one exevution uses one library)
also is detection of markers in the same image but from multiple dictionaries within the scope of this application (one execution multiple library)?

src/main/java/Application/OpenCVClient/MasterCommunicationRunner.java
```diff
@@ -0,0 +1,39 @@
...
+            try {
+                System.out.println("setup server");
+                activeOperator.setupTcpConnection("172.17.0.1:10200");
```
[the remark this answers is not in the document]
theodorelu27: I was testing and forgot to revert it back, I've fixed it with its appropriate value (from MasterCommunicator's getServer)

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
theodorelu27: at that time I only focused on getting it to work with existing code so I wrote the file to fit it into the existing system (getting mat with filename using imread)
I just researched and I think it wouldn't take long to use the buffered image, I'd try to change it to use the Buffered Image directly

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
theodorelu27: I put it as a parameter because beat command and get server command dependes on how it is implemented in the Master, so I wanted to give an access to change it without changing the code
I'm considering to add a TODO to move it to a config file

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
theodorelu27: it is an old TODO , for now it has been implemented by the Server throwing an IOException and continuing the loop

theodorelu27:
@alyhasansakr so for the feedback on runners, the TODO is to change the current implementation to not rely on while for looping?

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
alyhasansakr: yes, just add a TODO to move it to a config file

src/main/java/Application/Utilities/EdpHeartbeat.java
alyhasansakr: no problem

src/main/java/Application/Utilities/DetectMarker.java
alyhasansakr: just add a TODO here

alyhasansakr:
@theodorelu27 still use a while loop but with a public stop function, either implement it or write a TODO.

src/main/java/Application/OpenCVServer/OpenCVServerOperator.java
```diff
@@ -32,47 +29,59 @@
...
+            out = new PrintWriter(clientSocket.getOutputStream(), true);
+            in = new DataInputStream(clientSocket.getInputStream());
+            connected = true;
+        }
     }
 
     public void processing() throws IOException, IllegalArgumentException {
         currentImage = ImageIO.read(ImageIO.createImageInputStream(clientSocket.getInputStream()));
-        Mat subject = ImageProcessor.getBufferedImageMat(currentImage);
+        if(currentImage != null) {
```
theodorelu27, line 55: @alyhasansakr @juan-castrillon 
I'm having a problem here, when server starts before the client
if the client is connected and then fails, the server didn't immediately recognize that the connection is gone so it continues the loop with the BufferedImage being null, so when client tries to reconnect

the same does not apply when client is started first however, if Client started first and I forced it to stop, there's no problem in reconnection

I'm still researching why but any suggestions would be welcomed

alyhasansakr: @theodorelu27 Normally the server should be the one that always waits for the client, for example the function "accept" and "read" should block the server until a client is connected. The target is to make the server autonomous, regardless of the behavior of the client, the server must always work. And that includes recognizing that the client doesn't exist or doesn't respond.

Try one more time, if still can't do it, call me and we do it together.

theodorelu27: @alyhasansakr  I was able to reliably stop the connection when Client fails by sending an urgent message when the input stream returns null, thus throwing a Broken Pipe Exception when Client fails

so it's basically sending a byte as a heartbeat

alyhasansakr (approved):

Issue #22, opened 2020-08-16
Node-Master Interface: Sequence Diagram


alyhasansakr:
Develop a simple sequence diagram for the communication between a node and the master.
You may refer to the old code, search the Internet or just make your own.
This is a discussion topic, there is no right answer here.

Issue #4, opened 2020-07-28
Wiki Initial Content


alyhasansakr:
Add project description along with illustrative diagrams to the main page of the wiki.

juan-castrillon:
I started with the documentation as we talked on Monday @alyhasansakr .

If you check the wiki, here is the changes that i made:

- Added a home page with a description of the platform and a context diagram
- Created a general structure for the documentation: Filled some parts but the ones with a bullet point (- [ ] a) are still to be made

What do you think? Am I too detailed or not enough? Any feedback would be greatly appreciated.

alyhasansakr:
That is fantastic! just go on.

juan-castrillon:
As of right now, the whole [Using the platform](https://github.com/EdgeLab-FHDO/Edge-Diagnostic-Platform/wiki/Using-the-platform) section is almost complete. Only thing missing is the AdvantEDGE [example](https://github.com/EdgeLab-FHDO/Edge-Diagnostic-Platform/wiki/Running-with-AdvantEDGE#example), I wanted to do it with a real advantEdge scenario, generated in the platform, but I will have to wait for a bit because I am having some issues with my deployment and dont really seem to figure out a way to fix it, so I will do a clean install.

Any feedback would be greatly appreciated. I will start now with the [Basics](https://github.com/EdgeLab-FHDO/Edge-Diagnostic-Platform/wiki/Basics) section.

juan-castrillon:
Quick update, I was able to kind of fix my AdvantEDGE deployment. (Almost because it never turns green, always says `meep-postgis0: Pending`), therefore I could at least show some example in advantEdge client (Creating the scenario for real but not deploying)

juan-castrillon:
As of right now, the documentation is on the following:
- Basics section in the wiki is finished (I based myself  sometimes on what i am writing on the thesis)
- Using the platform section is also finished (With the current state of the platform)
- README now has a quick introduction and a link to the wiki

Still missing:
- Software Development part of the wiki: I am waiting on this according to what you told me @alyhasansakr about waiting on the modeling.

If we are to wait until the refactor to model, I would say we are ready to close this issue. Let me know what you think..

alyhasansakr:
That is more than enough for now.

PR #36, opened 2020-10-01
Interface Documentation

### LINKED ISSUE #21: Application-Master Interface: Sequence Diagram
Develop a simple sequence diagram for the communication between an application and the master.
You may refer to the old code, search the Internet or just make your own.
This is a discussion topic, there is no right answer here.

juan-castrillon:
I created the pull request, to add the documentation of the interfaces and resolve #21 and #22 

As a first step, I started with the App-Master Interface. There, I identified two scenarios to be modeled:

- Node Request Scenario
- Application Control scenario

This first commit, includes the first approach for the node_request scenario. The diagram as an image is:

node_request

It models the following:
- The client asks a master for a node, and the master gives one
- If the masters response was without errors, the client tries to connect to the server
- Once connected, and while running, the client constantly updates the master on its status with the server (CONNECTED or DISCONNECTED)
- If the Master realizes the Client is disconnected, sends a new node

Is still missing:
- **How to handle when the master doesnt give a "positive" response in the request**. I can think of the client asking again until one is given, or simply passing the error and exiting.
- **Dealing with an error response for connecting to the server the first time**. Same options as in the last one, I would say
- **Showing the messages as implemented with a REST API**. Not entirely sure if we want to keep the modeling independent from the implementation

What do you think? @alyhasansakr


alyhasansakr:
@juan-castrillon @theodorelu27 We discuss this next meeting.

juan-castrillon:
Based on what we discussed, this is the interface for the client and the master: 

node_request

- The threaded functionality we discussed is modeled with the par block
- If the client is not connected always executes locally, but tries to get a server
- The master receives a "assign_node" signal to realize the client exist and only gives him the node when asked

If you agree with this model I will get started with the Node-Master

juan-castrillon:
Here is the node-master interface model as we discussed:

Status_signal

- The server sends an asynchronous status (busy, free) signal to the master 
- The master acts on this information and also starts a timer
- If another status signal is not received before the timeout, the master assumes the server is down and deletes it from it node list

alyhasansakr:
I have been thinking about including match-making at all in these interfaces, maybe we should use the node-master diagram for both the servers and the clients, and remove the first part in the client (assign_node() and match_making()). Then match-making should be an independent task that is executed based on a group of conditions (one of which would still be a client joining the network).
@juan-castrillon @theodorelu27 What do you think?

juan-castrillon:
@alyhasansakr What benefit would it have to do it that way? Because the way I see it, and I may be wrong,  if the master runs the match making when a client asks (with assign_node() for example) it makes the system reactive in opposite of having to poll constantly to look if a new client joined the network no?

alyhasansakr:
@juan-castrillon 
We may need to run match-making in other cases as well, for example, when an edge node goes down, the master may need to do match-making again, maybe other conditions as well, we want it be as flexible as possible.

Another point that also requires that, we want to make a client and server evaluation system, which has the ability to drop a client or a server if they make problems (intentionally or unintentionally). Remember that in an actual deployment you may not have that much control over clients and servers. I don't want to make it complex in the future for features like this one. One of the requirements is that clients, servers and the master are all autonomous components that should always work regardless of how the others behave.

@theodorelu27 I also want to hear your opinion here.

juan-castrillon:
I see, then I agree with you, but in a way, the match making would still be triggered by a clients request right? It will be one of the triggers (not the only one). If so, I dont see the need to take it out of the modeling of the specific scenario. Maybe we can leave it as a form of clarification in the model? So it will be clear that a client can trigger match making (which is independent and internal to the master)?

alyhasansakr:
We will probably need another diagram for match-making, this one here should not mention it, so that it is not a rule here. Also I never regretted splitting functions.

Anyways, @theodorelu27  is the decider here, what do you say?

alyhasansakr (approved):
will be continued as part of #4

PR #40, opened 2020-10-13
Multiple Outputs Fix

alyhasansakr:
If one input goes to multiple outputs there is a chance that one of the outputs receive an invalid command, in such cases the for loop should not be terminated, instead it should just print the exception and go for the next input.


alyhasansakr:
This is not a good solution, we will design a better solution later.

juan-castrillon (approved):
Makes, sense, as it is right now, it would exit the loop, I also agree in that now that we are the ones running it, should be fine to leave it like this but eventually make it better

PR #42, opened 2020-10-19
Fixing deployAEScenario method. It was using wrong path of the API. N…

JeanValjean17:
…ow it takes the scenario name input and writes it into a json and post it to: /platform-ctrl/v1/sandboxes. This creates a new sandbox with a random generated name, and deploys the scenario.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -29,5 +29,7 @@ jobs:
       run: chmod +x gradlew
     - name: Open port 10500 for testing
       run: sudo ufw allow 10500
+    - name: Open port 4567 for testing
+      run: sudo ufw allow 4567
     - name: Test with Gradle ##Only testing with Gradle (Not building)
       run: ./gradlew test
--- a/src/main/java/InfrastructureManager/AdvantEdge/AdvantEdgeClient.java
+++ b/src/main/java/InfrastructureManager/AdvantEdge/AdvantEdgeClient.java
@@ -135,7 +135,7 @@ public class AdvantEdgeClient implements MasterOutput {
      * Terminate the current running scenario in AdvantEdge (Using the REST API)
      */
     private void terminateAEScenario() {
-        String requestPath = this.requestPath + "/active";
+        String requestPath = this.requestPath + "/sandbox-ctrl/v1/active/";
         //String requestPath = "https://postman-echo.com/delete";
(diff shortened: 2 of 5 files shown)


juan-castrillon (commented):
Hey @JeanValjean17 !

Is there any specific reason to why we need it wrapped in a sandbox, instead of just deploying the scenario? Because for just that, the path [was right](https://github.com/InterDigitalInc/AdvantEDGE/wiki/api-sandbox-ctrl#activename), it just required the scenario to be in the scenario store first (added).

Also, and this is because I really dont know it and maybe you have more experience, is it necessary to add the "/platform-ctrl/v1/" part ?, given that the API specification just mentions the path like "/sandboxes" or "/active"

alyhasansakr:
@juan-castrillon Sandboxes is a feature that was added recently, it is more generalized than having a single active scenario at a given time, so I would say we go with the sandboxes better.

About the exact path, of course you have to test it just to be sure, but you normally find it in the swagger file with the tag "basePath", that is why they may not mention it in a wiki page.

So try it and see if the new path is correct.

juan-castrillon:
@alyhasansakr You are right, so I will approve these changes and also then correct the other paths (for creating and terminanting scenarios) by adding the base URL

juan-castrillon (approved):

juan-castrillon:
My last commits are failing the test for something competely unrelated to what I modified: 

> InfrastructureManager.RestFunctionalTests > classMethod FAILED
    java.lang.Exception at RestFunctionalTests.java:24

I believe, it has something to do with the github worker, because locally, the test pass

alyhasansakr:
@juan-castrillon 
Not sure if this is the problem but it is a problem nevertheless, in the file RestFunctionalTests.java you are using getRestRunner() without parameters, that would throw an exception if there no RestRunner instance is created, instead use getRestRunner(String name, int port) that should create an instance in this case.

As for why it fails on the worker and not on your machine, if this is the problem then it has to do with the order of tests execution (if RestRunner was created before, it will not throw an exception here).

Fix that and lets see if it still fails.

juan-castrillon (approved):

juan-castrillon:
@alyhasansakr That was the problem, thanks a lot! Weird that it just pops up now, but now, the branch is ready to be merged, so i will do that

Issue #31, opened 2020-09-05
Match-Making


alyhasansakr:
Which edge node server is assigned to which application. Match-making should implement both MasterInput and MasterOutput. It receives application requests (usually as result of a REST POST call), and chooses a server and write it to a REST resource, which is read by an application REST GET call.

Currently, we only need the random algorithm, so server selection is random. But of course the implementation should allow the algorithm to be swapped (just by modifying configuration).

#21 #22 must be done before starting this one.

Don't forget the tests.

PR #43, opened 2020-10-20
Match making

juan-castrillon:
This is my first attempt to eventually solve #31 .

In this version the following was done:

- Created a `MatchMaker` class, as well as a `MatchMakingAlgorithm` interface (To be implemented by the different future algorithms) and a concrete implementation of this interface called `RandomMatchMaking`.

- Created simple classes to represent clients and nodes (`EdgeClient` and `EdgeNode`) . Right now, each has only basic fields (Client has only ID, and Node has ID, IP and status). These serve to convert from/to JSON files for example and facilitate their use as REST resources.

- Added a list of available nodes to the Master (Right now, filled with dummy nodes, for testing)

- POST request to the `/client/assign` URI, which contain the clients info as request body (JSON), make the `MatchMaker` assign them a node using the selected algorithm.

- GET requests to `/client/get_node/:client_id` are responded with the node information (the one previously assigned to that client) in form of a JSON object. (`:client_id` is a parameter)

- The testing I did was a manual workflow using postman, and can be seen in this GIF:

Vkpjh0I1SI

The issues and questions i have in order to continue are the following ones:

- The main issue, and I would appreciate some help with this one, is that right now, I dont have clear at all how to connect the MatchMaker to the Master. So in this version it works completely independent, but I tried to make it MasterInput or asterOutput and the functionality didnt make sense to me ( Would it be similar to a REST input, which sends a command to the master to activate itself as an Output?). I think i am not getting it right, thats why I did it independent first. Some clarification on that, would be nice @alyhasansakr 

- Automated testing is not yet done

Let me know what you think 😃


alyhasansakr:
@juan-castrillon Good job running it independently. But you can see the problem, if we now have a client that doesn't use "/client/assign" to register itself, or if we have a use case where the assigned nodes should be written to a file instead of "/client/get_node/:client_id".

To provide the needed flexibility, we need to implement match-making as both a MasterInput and a MasterOutput, the output part receives client and node registrations as well as match-making trigger, while the input parts produces the assigned nodes as commands.

You can just connect console input and output to test it, and when you deploy it you can change that to REST input and output, and so on.

Call me if you still have a problem with it.

juan-castrillon:
The last commit, already implements the Match Making as a MasterOutput and MasterInput.

I have made manual tests, including:

- MatchMaker MasterOutput part now handles register request for both nodes (`/node/register`) and clients (`/client/register`) and perform the match-making, then saves it to a `registeredClients` data structure in the master. This has been tested using postman POST requests.
- MatchMaker MasterInput part, will order the master to create the REST resource (containing information of the assigned node) for a specific client given its ID, and `RestOutput` will return this result to GET request made to `/client/get_node/:client_id`. This has been tested using console (writing the command directly) and then accesing the resources with postman GET requests

This is not finished yet, I only wanted to show where I am so far. Now the connection between the two is missing (as well as automated testing). For this, and as discussed previously with @alyhasansakr the configuration of the master has to be changed in order to guarantee that the same instance of MatchMaker is viewed as Input and Output in different threads (runners).

src/main/java/InfrastructureManager/Master.java
```diff
@@ -9,6 +12,9 @@
 
     private final CommandSet commandSet;
     private final ArrayList<Runner> runnerList;
+    private final List<EdgeNode> availableNodes;
+    private final Map<EdgeClient,EdgeNode> registeredClients;
```
alyhasansakr, line 16: while it will functionally work the same, it is better to have a separate list for clients, and rename this to clientNodeMapping.

juan-castrillon:
In the last commit, everything is working as expected, here is a summary of what was done since the last commit:

- The master was given an individual `EdgeClient` list, and the assignment of client and nodes is stored in a map called `clientNodeMapping`

- The configuration of the master was changed. Now, it has a field like:
```json
{
"ioData": {
      "inputs": [
        {"type": "ConsoleInput","name": "console_in"},
        {"type": "RestInput","name": "rest_in"},
        {"type": "MatchMaker","name": "match_m"}
      ],
      "outputs": [
        {"type": "ConsoleOutput","name": "console_out"},
        {"type": "RestOutput","name": "rest_out"},
        {"type": "MasterUtility","name": "util"},
        {"type": "AdvantEdgeClient","name": "advant_e"},
        {"type": "ScenarioDispatcher","name": "scenario_dispatcher"},
        {"type": "ScenarioEditor","name": "scenario_editor"},
        {"type": "MatchMaker","name": "match_m"}
      ],
      "ports": [
        {"name": "advant_e", "port" : 10500}
      ]
    }
}
```
This block, allows for configuration of output and input instances, by declaring the type and assigning each instance an unique name. So from here on in the configuration, each time an input or output is refered to, if it has the same name, will be considered the same instance in the different runners(MatchMaker for example)

- The new configuration was debugged and tested manually by using postman POST and GET requests:

1U9ltQpLxD

- Automated tests were added to test MatchMaker as an Output, Input and test the complete client flow (registering and then asking for node)

src/main/java/InfrastructureManager/Configuration/MasterConfigurator.java
```diff
@@ -0,0 +1,192 @@
...
+            this.inputInstances.put(inputData.getName(), getInputFromType(inputData.getType()));
+        }
+    }
+
+    /**
+     * Extracts output data from the configuration file, including ports
+     */
+    private void fillOutputInstances() {
+        for (IOConfigData outputData : this.data.getIoData().getOutputs()) {
+            if (this.inputInstances.containsKey(outputData.getName())) { //If an instance is input and output
```
alyhasansakr, line 69: Why not the same for fillInputInstances? You don't have a problem because of the call order (fillInputInstances before fillOutputInstances), but I don't see a condition that defines the order of calling fillInputInstances and fillOutputInstances.

src/main/java/InfrastructureManager/Master.java
```diff
@@ -175,9 +226,13 @@ public static Master getInstance() {
         return instance;
     }
 
+
+
     public static void main(String[] args) {
         Master.getInstance().startMainRunner();
+        Master.getInstance().startRunnerThread("RestInputTest");
         Master.getInstance().startRunnerThread("RestServer");
+        Master.getInstance().startRunnerThread("MMInputTest");
```
alyhasansakr, line 235: no longer needed

src/main/java/InfrastructureManager/MatchMaking/MatchMaker.java
```diff
@@ -0,0 +1,84 @@
...
+
+    private void assign (EdgeClient client) {
+        EdgeNode node = this.algorithm.match(client);
+        if (node == null) {
+            System.out.println("No node to assign");
+        } else {
+            Master.getInstance().mapClientNode(client, node);
+            command = "give_node " + client.getId();
+        }
+        System.out.println("Client received : " + client.getId() + " Server assigned : " + node.getId());
```
alyhasansakr, line 31: no longer needed

src/main/java/InfrastructureManager/MatchMaking/MatchMaker.java
```diff
@@ -0,0 +1,84 @@
...
+                throw new IllegalArgumentException("Arguments missing for command - MatchMaker");
+            }
+        }
+    }
+
+    private void registerClient(String clientAsString) {
+        try {
+            EdgeClient client = this.mapper.readValue(clientAsString, EdgeClient.class);
+            Master.getInstance().addClient(client);
+            assign(client); //Perform MatchMaking
```
alyhasansakr, line 59: assigning a node to a client should be a separate command, remove it from here and add a new command to out().

src/main/java/InfrastructureManager/MatchMaking/MatchMaker.java
```diff
@@ -0,0 +1,84 @@
...
+    public void setAlgorithm(MatchMakingAlgorithm algorithm) {
+        this.algorithm = algorithm;
+    }
+
+    private void assign (EdgeClient client) {
+        EdgeNode node = this.algorithm.match(client);
+        if (node == null) {
+            System.out.println("No node to assign");
+        } else {
+            Master.getInstance().mapClientNode(client, node);
```
alyhasansakr, line 28: this coupling prevents us from running multiple instances of match-making, it would be better to have each instance of match-making be self-contained. Basically remove master use in MatchMaker and instead use lists as instance variables.

src/main/resources/Configuration.json
```diff
@@ -1,22 +1,47 @@
...
+        {"type": "ConsoleOutput","name": "console_out"},
+        {"type": "RestOutput","name": "rest_out"},
+        {"type": "MasterUtility","name": "util"},
+        {"type": "AdvantEdgeClient","name": "advant_e"},
+        {"type": "ScenarioDispatcher","name": "scenario_dispatcher"},
+        {"type": "ScenarioEditor","name": "scenario_editor"},
+        {"type": "MatchMaker","name": "match_m"}
+      ],
+      "ports": [
+        {"name": "advant_e", "port" : 10500}
```
alyhasansakr, line 18: that looks unnecessarily complex, why not just add the port to the output line? JSON is flexible in this regard.

src/main/java/InfrastructureManager/Master.java
juan-castrillon: I believe it is needed because that is the thread that has MatchMaker as input, so is the one that serves the clients with the REST resource. I could maybe change the names to make that more clear (get rid of "test")

src/main/java/InfrastructureManager/MatchMaking/MatchMaker.java
juan-castrillon: Couple of questions here: 
- Do you mean remove all Master usage? Including for example `addClient` or `addNode` methods? Or just this one specifically?
- How would the replacement with lists look like? Could you maybe give me an example?

alyhasansakr: Yes including addClient and addNode.

Here you are keeping 3 lists: clients, nodes and client-node mappings. These lists should be instance variables for each instance of the MatchMaker.

The Master should have at least 2 lists: clients and nodes (whether it has a mapping is up to you). For now you can use the same definition for client and node in the MatchMaker and the Master, but at some point it will be different. What am trying to say is that the Master sees the client from a different point of view than the MatchMaker, and thus needs to be independent.

For now, anytime a client is added to the Master it should also be added to the MatchMaker, we will need to change that later for specific uses cases, I just don't want to have a structure that is not flexible.

src/main/java/InfrastructureManager/Master.java
alyhasansakr: Now that I think about it, even RestServer is not needed here, you can always start them when you are need them, but it should be possible to not start them if the use case doesn't include a rest server or match-making.

juan-castrillon:
Ok, in the last commits, I got rid of any coupling between match making and master (including the one in the match making algorithm) and assigned the client and node registration in the master to the restInput (Which then sends the order to matchmaker). 

The manual test with postman was sucessfull and I also refactored the automated test for the new code

Let me know what you think

juan-castrillon:
Now its complete

src/main/java/InfrastructureManager/Rest/RestOutput.java
```diff
@@ -87,4 +96,18 @@ public void resetOutput() {
         this.output = new LinkedList<>();
         this.limitNodes = new LinkedHashMap<>();
     }
+
+    private void createNodeResource(String clientID, String nodeID) {
+        try {
+            EdgeClient client = Master.getInstance().getClientByID(clientID);
+            EdgeNode node = Master.getInstance().getNodeByID(nodeID);
+            if (node == null) {
+                throw new Exception("Client not found");
```
alyhasansakr, line 105: is that a client or a node? why not for both?

src/main/resources/OldConfiguration.json
```diff
@@ -0,0 +1,54 @@
+{
```
alyhasansakr, line 1: delete old configuration

src/main/java/InfrastructureManager/Rest/RestOutput.java
juan-castrillon: My bad, this is from the old implementation. This is not even needed as both methods `getClientByID` and `getNodeByID` cannot return null and already throw an exception when the resource is not found. The last commit already has this fixed.

alyhasansakr (approved):

Issue #41, opened 2020-10-14
Master Multi-output Separation


alyhasansakr:
Currently the configuration specifies the following independently:
- Which input connects to which output.
- Mapping between input commands and output commands.

The problem is the following: You may need to have a single input command that is mapped to two different output commands for two different outputs.

Currently this is possible by just adding the two different mappings, however, this means that both outputs are going to receive ALL output commands, which includes output commands that are not defined for them.

For example, you have a single event that should limit both bandwidth and CPU, but bandwidth output is different than CPU output. Using what we have now, bandwidth output commands have to go to CPU output as well, which throws an exception.

Possible solution: The configuration should specify input, output and command mapping in the same rule.

For example, rules that map event input (from scenario S1) to bandwidth limiter (BL1) and CPU limiter (CL1) should look like the following:
```JSON
{
   "rules":
   [
      {
         "input": "S1",
         "output": "BL1",
         "inputCommand": "applyLimits $bandwidth_limit $CPU_limit",
         "outputCommand": "advantEdge networkUpdate sandbox_1 edge_1 EDGE $bandwidth_limit $bandwidth_limit 2 1 0.001"
      },
      {
         "input": "S1",
         "output": "CL1",
         "inputCommand": "applyLimits $bandwidth_limit $CPU_limit",
         "outputCommand": "limit edge_1 $CPU_limit 100000"
      }
   ]
}
```

For that to work correctly, we need to define inputs and outputs in the configuration file (like "rest": "CL1").

You can also notice that there is one more requirement here: Parameters mapping. Commands have parameters which needs to be mapped as well. My suggestion is to use the symbol $ to indicate that whatever comes here in the input command is copied to the corresponding one in the output command. For that I don't want to implement a pattern matching functionality to extract the parameter values from the input, instead just enforce a rule for separation using spaces (as we already have), which means that the following is not a valid input command "applyLimits ($bandwidth_limit,$CPU_limit)" since it doesn't separate with spaces. This rule is not enforced for the output command replacement, because it is just a trivial find and replace function.

juan-castrillon:
I have a couple of questions and want to clarify everything before getting on this issue:

The requirements are:

- Configuration of the master should allow to specify for a given input command (e.g applyLimits) a mapping to not only one (like currently) but more outputs.

[...]

PR #49, opened 2020-10-29
Multi out config fix

juan-castrillon:
First attempt at solving #41 . In all the commits a lot of changes were made, the main ones being:

- The configuration file was modified according to what was discussed in #41 
- Runner creation is now made in a way, that a runner is created for each of the declared inputs that has a connection.
- Scenarios and the ScenarioRunner creation, are also handled as an input, using the name of the scenario as the input name.
- MasterOutput is now an abstract class. instead of an interface, because it provides a field to store the instance name of an instance, this in order to identify it when passing commands to it. All MasterOutputs then now have a constructor where this name is set.
- The interfaces MasterConfig and MasterResource were deleted
- Command set is no longer a singleton
- Master no longer has a single CommandSet but instead, each runner has a CommandSet for each of the outputs of a given input (The one for which the runner was created). This is used when running master.execute therefore assuring the defined commands go to the defined output instances.
- The main runner is not defined as the runner that has "console_in" as input

Right now, these things missing are:
- Changing the name of the new config file and deleting the old one
- Parameter Mapping (Is working right now like it always has, just copying whatever it has after the defined input command to the output command)
- Fixing the automated test that depend on the old configuration (all of the MasterMapping tests, one for Match Making and the NodeLimit). However, the other tests still pass and I did run some manual tests to check the matchmaking (postman) and the scenarios and they are working fine. 

Let me know what you think @alyhasansakr


juan-castrillon:
In the last commits:

- I added parameter mapping, specifying in the configuration file in the following way for example:
```json
{
      "in" : "console_in",
      "out" : "rest_out",
      "commands" : {
        "limit_node $tag $CPU_limit" : "restOut limit $tag $CPU_limit"
      }
}
```
copies the parameters.
- Got rid of unused imports
- Changed the naming of the RestOutput singleton which was causing the test to fail in the GH worker. Now, the `getInstance()` method is used and no name for the RestOutput is specified (right now this only happens when testing), the name sets to "rest_out" as default.

Missing are:

- Changing the config file name and deleting the old one
- Redoing the MasterMapping tests (Which I dont know if we really need?)

alyhasansakr (approved):
we merge this one now, and refactor it later

Issue #38, opened 2020-10-11
General Terminal Command Execution


alyhasansakr:
Implement a MasterOutput that allows execution of terminal commands on a remote machine given its ip address (and port if necessary). Currently, we will only support BASH on Linux.

As a first step, we don't care about the return of the command (whether it was successful or not).

To achieve that, you may use one of the following or come up with your own solution:
- SSH.
- BASH script (on the remote machine) that starts a server to accept commands.
- Java application (on the remote machine) that starts a server to accept commands.

Include unit tests for both the masterOutput and any other application or script you develop (if any) as well as a manual end-to-end test.

PR #65, opened 2020-11-25
Remote Terminal Execution

juan-castrillon:
Feature that will eventually solve #38 

What I have done so far is the following:
- Implemented a solution for the issue with SSH (no aditional components) using [JSch](http://www.jcraft.com/jsch/)
- Created a new `MasterOutput` called `SSHClient`
- Added two control commands: `execute` and `setup`
- In the `execute` command, added a flag to determine if the command is to be run in background in the remote machine or if the output is waited for and printed on screen
- The `setup` command allows to configure the SSH connection with: Host, Port, Username and Password
- Maped the output  in the configuration to some basic commands
- Tested basic commands, connecting to a VM running Ubuntu 18
- In the same VM, created the following dummy script:
    ```bash
    #!/usr/bin/env bash
    while :
    do
      echo "Doing something"
      sleep 2
    done
    ```
- Executed the dummy script from the Master, without hanging on completion (on the background)
- Added the script to `src/test/resources/Manual Test Resources/dummy.sh` for future use

Still missing are:
- Support for running `sudo` commands in the remote machine
- Automated Tests
- This one is optional, but i thought I would be a good idea to add support for sending files to the remote via SCP (The library has support) what do you think @alyhasansakr


alyhasansakr:
@juan-castrillon Adding an option to send files is really good, do it as well.

juan-castrillon:
In the last commits I:

- Added support for sudo commands (both in front and background)
- Added SCP protocol implementation for sending from the master to the remote
- Modified the way the IO resources are handled to avoid leaks
- Added automated tests (To an extent, just tests exceptions)
- Added instructions and resources for running a manual test. This can be seen in `src/test/resources/ManualTestResources/SSHClientTest/`

I believe is ready for review

alyhasansakr (approved):

Issue #54, opened 2020-11-13
File Writer Output


alyhasansakr:
Implement a MasterOutput that writes to a file, you are free on how to design it, but it should be able to create a new file in runtime.

Don't forget the tests.

PR #67, opened 2020-12-02
File Output

juan-castrillon:
Pull request to solve #54 

In these commits I:

- Created a `MasterOutput` called `FileOutput` and added it to the config file and the configurator
- Added a functionality to create a file based on a path and content (as a string, can include spaces) (This always overwrites)
- Added a functionality to change the enconding of the output file. If not set, the default was set to UTF-8
- Added a functionality to append to an existing file instead of overwriting
- Added automated tests that show all the functionalities working

After I fix the issue of the tests passing locally but not in the GH Workers, I believe is ready for review


juan-castrillon:
Now is ready

alyhasansakr (approved):

Issue #66, opened 2020-12-01
Match-Making Comparison Experiment


alyhasansakr:
The requirement is to run the following setup:
- Multiple client and server application instances running on AdvantEDGE locations (UE - FOG - EDGE).
- Run match-making to connect clients with servers
- Limit bandwidth using AdvantEDGE.
- Limit CPU using Cgroups.

First you will need to write a configuration file and a scenario file, running the scenario file should simply run the whole experiment.

The output of this issue is to write a wiki page of how to do this experiment.

PR #74, opened 2020-12-22
Matchmaking Experiment

theodorelu27:
Pull Request for required changes to run the experiments for #66


theodorelu27:
@alyhasansakr the first draft of the wiki is done at https://github.com/EdgeLab-FHDO/Edge-Diagnostic-Platform/wiki/Demo:-Marker-Detection-Application

as discussed, all changes related to the experiments are included in this pull request

src/main/bash/NodeLimit/limit_docker.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
```
alyhasansakr, line 1: Why do we need this file?

theodorelu27: I forgot to remove this file entirely, I previously made another file since container cannot run sudo, but as we wouldn't run it in the container, it is no longer needed

alyhasansakr (approved):

PR #78, opened 2021-01-07
Modules refactor v1

### LINKED ISSUE #69: Runner Refactoring: Modules
There is no new functionality needed here, it is just a better way to implement cases similar to match making, that is when you want to have more than a single input or output access the same instance.

To properly define that, we should not define inputs and outputs in the configuration file, instead, we define modules. Each module has a predefined set of inputs and outputs which can still be used the same way in connections (with added prefix to indicate which instance of the module is used). For example, we define a match making module with the name mm (and some parameters), then we can use mm.in as an input ("in" is a hard coded name) and mm.out as an output that will access the same instance. It may be that a single module has more than one input each with a unique name and functionality, but all of them access the same module instance.

This also help us defining namespaces and tests, as each module should have its own namespace (and folder) and set of tests.

Don't forget to write an abstract class Module and use it for each module we already have (console, rest, scenario, etc).

juan-castrillon:
Pull request to eventually close #69.

As of right now, I just creating the request as we discussed to show my progress but the whole thing is still messy.

Right now, Is kinda working with only one module and no exit or pause capabilities have been tested yet.

So far:
- I fixed a package name issue with the new REST package (Name was converted to lowercase for some reason)
- Added autostart as a console argument as we discussed the last pull request
- Created `MasterInput` as an abstract class
- Added classes necesary for modules (Abstract class, Manager, Factory,Connector) as well as some enums and exceptions
- Modified configuration and runners

Of course, tests will fail


juan-castrillon:
@alyhasansakr  I could use some help or a fresh perspective here.  The refactor, as you can kind of see in the commits names, is working, and altough many things still have to be changed, I am close to being able to get rid of Runners and ScenarioRunners to start then modifying each individual module.

To do this, the scenario module was implemented, however with this piece of code (also present before) i have an issue:
```java
private void waitForEvent (Event e) {
        long absoluteTime = this.getStartTime() + e.getExecutionTime();
        try {
            Thread.sleep(absoluteTime - System.currentTimeMillis());
        } catch (InterruptedException ie) {
            ie.printStackTrace();
        }
    }
```
I dont know why this wasnt an issue earlier but if the scenario is paused, when resumed, the substraction  `absoluteTime - System.currentTimeMillis()` is of course negative and therefore doesnt work. It should be kind of easy but I can think of a new way to guarantee the timing, do you have any ideas?

zero212:
> @alyhasansakr I could use some help or a fresh perspective here. The refactor, as you can kind of see in the commits names, is working, and altough many things still have to be changed, I am close to being able to get rid of Runners and ScenarioRunners to start then modifying each individual module.
> 
> To do this, the scenario module was implemented, however with this piece of code (also present before) i have an issue:
> 
> ```java
> private void waitForEvent (Event e) {
>         long absoluteTime = this.getStartTime() + e.getExecutionTime();
>         try {
>             Thread.sleep(absoluteTime - System.currentTimeMillis());
>         } catch (InterruptedException ie) {
>             ie.printStackTrace();
>         }
>     }
> ```
> 
> I dont know why this wasnt an issue earlier but if the scenario is paused, when resumed, the substraction `absoluteTime - System.currentTimeMillis()` is of course negative and therefore doesnt work. It should be kind of easy but I can think of a new way to guarantee the timing, do you have any ideas?

Hi Juan, Zero here
Can you tell me what exactly waitForEvent here do? wait for (x) milisecond base on what event is in the parameter?

Then shouldn't it look like this 
```java
try {
         Thread.sleep(e.getExcecutionTime());
        } catch (InterruptedException ie) {
            ie.printStackTrace();
        }
}
```

juan-castrillon:
Hey Zero!

You got it right, with this we assure that the event execute in the correct time. 


In the first iteration of Scenarios we decided that with this way (as opposed to what you mention), we could guarantee the execution in certain points in time, while accounting for little delays inside the code. So for example if you want an event in precisely time instant X then with this way, you guarantee it will happen in X after the scenario started. However it fails when the scenario is paused

alyhasansakr:
@juan-castrillon 
One way is to add pause time to the calculation. Pause time is the summation of time differences between each pair of pause and resume.

Example:
- start = 10
- paused at 15
- resumed at 20
- next event is 30
- should wait until = start + (resumed - paused) + next event = 10 + 5 + 30 = 45

Keep in mind that (resumed - paused) is calculated on the pause and resume functions and is accumulated. Stopping should reset this value to zero.

zero212:
I still think this is an overkill for what you are trying to implement, I guess you want 100% accuracy of when the next event will start, to avoid situation like 

- start time = 0, 
- wait for event A (2000ms) - 
- at 2001 - start event B. 

But due to little delay by java computation, wait for event A may not sleep for exact 2000ms, but 2004ms, hence you miss the event B?

Well then look at the body of your function
```java
long absoluteTime = this.getStartTime() + e.getExecutionTime();
        try {
            Thread.sleep(absoluteTime - System.currentTimeMillis());
        } catch (InterruptedException ie) {
            ie.printStackTrace();
        }
```
The only thing that change when scenario paused is System.currentTimeMilis(); 
I would assign a new variable call `pausedTime` where when scenario is paused, we assign `pausedTime = system.currentTimeMilis()` and when another one called `resumedTime` when resumed with similar manner. 
With that we can assured how much time has passed since we have paused the scenario. Then just add to the calculation

alyhasansakr:
@zero212 That is exactly what we are saying, don't worry we will get it right this time.

zero212:
@alyhasansakr lol I didn't see your comment when I was typing mine. but anyway I hope our comments are helpful to him :))

juan-castrillon:
@alyhasansakr @zero212 Thank you guys. Your comments really help. It seems you both propose the same idea and looks like it would solve my mental block 😆  I will start working on it 😄

juan-castrillon:
@alyhasansakr So far:

- I've created all classes in `ModuleManagement` package which include 
    - a `ModuleFactory` class that creates and instanciates the correct module according to the definition in the configuration and a defined "type" variable which is an enum. 
    - a `ModuleConnector` that creates `Connection` elements that are then passed to each module to in turn pass to each of the runners of its inputs.
    - a `ModuleManager` which contains the two.
    - an `PlatformModule` abstract class which has all the required functionalty for modules
- Modified the `MasterConfigurator` to use only the `ModuleManager` to create the modules
- Added a new  class in `Configuration.RawData` to account for the definition of a module in the config file, and uses Jackson to define inheritance to different modules based on the `type` parameter
- Created an tested the following modules:
    - Console
    - Utility
    - Scenario
    - REST

I will proceed with creating the other modules

juan-castrillon:
In the last commits, the only module left is MatchMaking. I am currently working on separating input and output by creating a shared resource with java's [ConcurrentMap](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentMap.html) with a Semaphore to block. I did the same for the other case of IO in one (NodeLimit). 

Missing are of course test and documentation, as well as maybe better exception handling, however, I wanted to finish first because of possible changes after review

juan-castrillon:
@alyhasansakr In theory, if this needs to be merged quickly, all the modules are already implemented and manually tested by me. Of course, like I said before, automatic testing is completely missing, as well as documentation and better exception handling. 

However, if you could start to have a look in the code, i can start making the necessary changes to then start documenting and creating and handling exceptions. 

In the meanwhile I will start with the testing

juan-castrillon:
As of right now, the functional part already has all the improvements we discussed, and exception handling is done, so missing are the automatic tests and documenting the code

juan-castrillon:
@alyhasansakr Apart from documentation, everything is ready and functional. I believe now is a good point for a review and at the same time I can start creating javadocs

juan-castrillon:
@alyhasansakr All requested changes are made, like we discussed, no javadocs yet but is ready for merging and I can open a new issue for the documentation (both code and wiki) and self-assign

alyhasansakr (approved):
Looks fantastic!

@juan-castrillon you can open a new issue for documentation.

PR #79, opened 2021-01-16
Network structure

lkh5kor:
- Added folder network structure as part of [InfrastructureManager](https://github.com/EdgeLab-FHDO/Edge-Diagnostic-Platform/tree/master/src/main/java/InfrastructureManager) with the required classes and few functions.

The first two commits contains basic class definitions. Range check for latitude, longitude and distance calculation from latitude and longitude which would be useful later. 

Pending items:
1. Relation classes implementation.
2. JSON string read and write implementation.
3. Unit testing.
4. Addition of comments to all files.

Creating pull request so as to start the review and also to discuss further steps.


alyhasansakr:
@lkh5kor Is it ready for review? you requested my review but I still see a lot of TODO's in the code.

lkh5kor:
> @lkh5kor Is it ready for review? you requested my review but I still see a lot of TODO's in the code.

@alyhasansakr , The complete implementation is not done. As per our previous discussion, I have created the pull request. Few of the todo are related to the additional functions like distance calculation using latitude, longitude co ordinates, range checks for latitude and longitude values... etc. which we decided to keep it as it is and later move them to different files/classes.
JSON string read and write and relationship between classes are pending.

alyhasansakr:
@lkh5kor Creating a pull-request is not the same as requesting a review. You should only request a review when you are done with the implementation and the code is ready to be reviewed.

lkh5kor:
@alyhasansakr ,
As per our discussion, I have updated the changes, unit testing is also done. 
Documentation/Comments are also added.
The pull request is now ready for review.
Kindly review 

Regards,
Shankar

src/main/java/InfrastructureManager/NetworkStructure/Network.java
```diff
@@ -0,0 +1,182 @@
...
+		deviceLocationList = new ArrayList<>();
+	}
+
+	/**
+	 * Function to read JSON string and update network class
+	 * @param String
+	 */
+	public void loadNetwork(String readString) throws JsonProcessingException{
+		
+		//New object created as the called object cannot be accessed as a whole. Therefore new object is created to update the calling object members with values from json string
```
alyhasansakr, line 44: it is better to write a copy function here.

src/main/java/InfrastructureManager/NetworkStructure/Network.java
```diff
@@ -0,0 +1,182 @@
...
+	/**
+	 * Function to save JSON string
+	 * @return String
+	 */
+	public String saveNetwork() throws JsonProcessingException{
+		String json = null;
+		try {
+			// convert network object to JSON
+			json = this.mapper.writeValueAsString(this);
+		} catch (Exception ex) {
```
alyhasansakr, line 64: why are you both handling and throwing exceptions at the same time?
here you shouldn't handle exceptions.

src/main/java/InfrastructureManager/NetworkStructure/Network.java
```diff
@@ -0,0 +1,182 @@
...
+		}
+		return json;
+		
+	}
+	
+	/**
+	 * Getter function
+	 * @return deviceList
+	 */
+    public List<Device> getDeviceList() {
```
alyhasansakr, line 75: I would advice against a full access to the device list, instead, a function that gets a device by its id makes more sense (mutable or not)

src/main/java/InfrastructureManager/NetworkStructure/Network.java
```diff
@@ -0,0 +1,182 @@
...
+	 */
+    public List<DeviceLocationRelation> getDeviceLocationList() {
+        return deviceLocationList;
+    }
+    
+	/**
+	 * Setter function
+	 * @param deviceList
+	 */
+    public  void setDeviceList(List<Device> deviceList) {
```
alyhasansakr, line 131: setting the whole list is not recommended, instead, adding and removing a specific device should be implemented.

src/test/java/InfrastructureManager/NetworkStructure/NetworkTest.java
```diff
@@ -0,0 +1,49 @@
...
+	@Test
+	public void saveNetworkTest() throws JsonProcessingException {
+	 // create network object
+    	Network network = new Network();
+    	network.setDeviceList(Arrays.asList(new Device(001),new Device(002),new Device(003),new Device(004)));
+    	network.setConnectionList(Arrays.asList(new Connection(100),new Connection(101),new Connection(102),new Connection(103),new Connection(104)));
+    	network.setLocationList(Arrays.asList(new Location(500),new Location(501),new Location(502),new Location(503),new Location(504)));
+    	network.setApplicationList(Arrays.asList(new ApplicationInstance(600),new ApplicationInstance(601),new ApplicationInstance(602),new ApplicationInstance(603),new ApplicationInstance(604)));
+		String writeString = network.saveNetwork();
+		System.out.println(writeString);
```
alyhasansakr, line 26: no need to print anything while testing, this is automatically done when a test fails.

src/test/java/InfrastructureManager/NetworkStructure/NetworkTest.java
```diff
@@ -0,0 +1,49 @@
...
+		int deviceID0 = network.getDeviceList().get(0).getDeviceId();
+		System.out.println(deviceID0);
+	    Assert.assertEquals(1, deviceID0);
```
alyhasansakr, line 29: you only tested the first device id value, what about all of the others values that you assigned?
and what about writeString?
also, you are mixing between testing the setters and testing saveNetwork, each should has its own test.

src/test/java/InfrastructureManager/NetworkStructure/NetworkTest.java
```diff
@@ -0,0 +1,49 @@
...
+		 // create network object
+		//ObjectMapper mapper = new ObjectMapper();
+    	//Network network = mapper.readValue(jSON, Network.class);
```
alyhasansakr, line 40: don't leave commented code without a reason.

src/test/java/InfrastructureManager/NetworkStructure/NetworkTest.java
```diff
@@ -0,0 +1,49 @@
...
+		Network network = new Network();
+    	network.loadNetwork(jSON);
+		int deviceID1 = network.getDeviceList().get(1).getDeviceId();
+	    Assert.assertEquals(2, deviceID1);
```
alyhasansakr, line 44: and what about everything else you included in the json string?

alyhasansakr (changes requested):
some of the comments are written only once but apply for more than one case, always check similar cases in your code.

src/main/java/InfrastructureManager/NetworkStructure/ComputeLimits.java
```diff
@@ -0,0 +1,86 @@
...
+	public int getMemoryLimit() {
+		return memoryLimit;
+	}
+	
+	/**
+	 * Setter function
+	 * @param cpuCount
+	 */
+	
+	public void setCpuCount(int cpuCount) {
```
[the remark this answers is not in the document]
lkh5kor: Wouldn't the setters be required at a later stage when you allocate cpu, gpu and memory for application and when you free them up after the application is completed.?

src/main/java/InfrastructureManager/NetworkStructure/Connection.java
```diff
@@ -0,0 +1,63 @@
...
+	 */
+	public NetworkLimits getNetworkCapacity() {
+		return networkCapacity;
+	}
+	
+	/**
+	 * Setter function
+	 * @param networkCapacity
+	 */
+	public void setNetworkCapacity(NetworkLimits networkCapacity) {
```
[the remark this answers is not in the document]
lkh5kor: Wouldn't the setters be required at a later stage when a new connection is established or existing connection is disconnected.? For requirements we may not need the setters as it is fixed but here don't we need the setters?

src/main/java/InfrastructureManager/NetworkStructure/NetworkLimits.java
```diff
@@ -0,0 +1,83 @@
...
+	 */
+	public double getPacketLoss() {
+		return packetLoss;
+	}
+	
+	/**
+	 * Setter function
+	 * @param latency
+	 */
+	public void setLatency(int latency) {
```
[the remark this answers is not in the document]
lkh5kor: Wouldn't the setters be required at a later stage when a new connection is established or existing connection is disconnected.?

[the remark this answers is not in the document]
alyhasansakr: that would be a part of a different implementation, and it will not use a setter in this way.

src/main/java/InfrastructureManager/NetworkStructure/ComputeLimits.java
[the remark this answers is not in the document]
alyhasansakr: part of future implementation.

src/main/java/InfrastructureManager/NetworkStructure/Connection.java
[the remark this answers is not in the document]
alyhasansakr: part of future implementation.

lkh5kor:
@alyhasansakr ,
The review findings are fixed.
Read from JSON String test case(loadNetworkTest) is commented as the string length is exceeding the max size required for Jackson serialization, otherwise all the other test cases are covered and assertion checks have passed.

lkh5kor:
@alyhasansakr ,
The review findings are fixed.
Read from JSON String test case(loadNetworkTest) is commented as the string length is exceeding the max size required for Jackson serialization, otherwise all the other test cases are covered and assertion checks have passed.
Sorry, the PR was closed by mistake, I have reopened it.

alyhasansakr:
@lkh5kor that is really good, there are still a few things that we are going to improve along the way.
Now you need to merge the master branch into this branch and write a new class NetworkModule that extends PlatformModule. The new module should have an output that creates a new network and an input that gives the current network.

juan-castrillon:
@lkh5kor Documentation is still needed for the modules, however to start follow this:

# How to create a new module

1. In the folder `src/main/java/InfrastructureManager/Modules` create a new folder with the module's name
2. Inside the new package create the module class (extending `PlatformModule`) and a `RawData` package with the raw representation of the module (`extends ModuleConfigData`).
3. In `src/main/java/InfrastructureManager/ModuleManagement/RawData/ModuleConfigData.java` add the new raw data class as a json subtype
4. In `src/main/java/InfrastructureManager/ModuleManagement/ModuleFactory.java` add a new type to the `ModuleType` enum
5. In the raw representation add the new type as the type parameter
6. Again in the factory handle the new case in the switch expression to return a new Module

lkh5kor:
@alyhasansakr, @juan-castrillon,
I have made the requested changes as per my understanding. Kindly review.
I am not aware of the current exception handling of modules. I have modified the src\main\java\InfrastructureManager\ModuleManagement\ModuleInput.java to handle one of the JSON exceptions as it was giving me error. I think that's not the proper way to handle the exception. 
Kindly let me know how it is to be handled.

src/main/java/InfrastructureManager/Modules/NetworkStructure/Input/NetworkInput.java
```diff
@@ -0,0 +1,28 @@
...
+public class NetworkInput extends ModuleInput {
+	
+	private final Network network;
+	
+	public NetworkInput(PlatformModule module,String name,Network network) {
+		super(module,name);
+		this.network = network;
+	}
+    @Override
+    public String read() throws JsonProcessingException  {
```
alyhasansakr, line 21: here only InterruptedException could be throw. JsonProcessingException or any other exception must be handled in this function.

src/main/java/InfrastructureManager/ModuleManagement/ModuleInput.java
```diff
@@ -35,6 +36,6 @@ public String getName() {
         return name;
     }
 
-    public abstract String read() throws InterruptedException;
+    public abstract String read() throws InterruptedException, JsonProcessingException;
```
alyhasansakr, line 39: remove JsonProcessingException

src/main/java/InfrastructureManager/ModuleManagement/ModuleOutput.java
```diff
@@ -1,6 +1,9 @@
 package InfrastructureManager.ModuleManagement;
 
 import InfrastructureManager.ModuleManagement.Exception.Execution.ModuleExecutionException;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
```
alyhasansakr, line 5: remove JsonProcessingException

alyhasansakr (approved):

PR #85, opened 2021-02-03
Requested Fixes, match making module (and extras)

juan-castrillon:
In this pull request are the requested changes for the MatchMakingModule as well as other issues identified since the modules refactor.

So far:
- Each match making algorithm is now a separate module. Their type is specifically defined in the sense that to configure for example a random match making module in the config file:
    ```json
    {"type" :  "MatchMakingModule-Random", "name":  "mm"}
    ```

- There is now an abstract class called `LoggableObject`. Any object in the code that extends this class, will need to be passed an owner module in the constructor and will have access to its `DebugInput` to log messages.

- There is an immutable version of the `PlatformModule` class, which is the one passed to inputs, outputs and loggable objects. It is a superclass of the original platform module class, which now only exposes the setters.

- `ModuleInput` is now called `PlatformInput` to show that is an input to the system, not a module. Same thing with `ModuleOutput` which is now `PlatformOutput`.

- `ScenarioEditor` now makes more sense, there is no `create` or `fromFile` functions anymore, given that each editor belongs to a module which has only one scenario.

Missing is:
- Right now, raw data classes of all match-making modules look the same. The configuration specific to each module (what should be different in each module in the config file) is still not 100% clear to me.
- That is also the case for IO creation. Again, some clarifying in the differences would be great.

- I am not sure if `LoggableObject` as an abstract class is the best idea. In my head, an interface would be better, but it needs to have a field (of `ImmutablePlatformModule`) in order to work. Is there any way to do it that I am missing?. The thing I dont like about the abstract class is that a child of another superclass could not be loggable in the future (No multiple inheritance allowed)

- Once that is solved, I will make the "example" of using the debug input in one of the match-making classes

If you guys could take a look @zero212 @alyhasansakr  would be great 👍


zero212:
I will have a look...but i don't think my knowledge about the modules is not sufficient to make any statement here. Would you mind having a meeting with me tomorrow so you can teach/show me how things work? 
Or maybe a `module-example` on how to create a module would be great?  Since now several classes/modules need to call other classes to use this and that (e.g. the LoggableObject)

alyhasansakr:
@juan-castrillon @zero212 We can have a meeting, but let's schedule on Slack.

juan-castrillon:
I did what we discussed last time and implemented the `PlatformObject` class, and made everything I think should extend it do it. However, I am stuck because this test:
```java
@Test
    public void disconnectAndChangeInScoreTest() throws Exception {
        matchMaker.execute("matchMaker assign_client client1");
        //job failed -> client 1 score with node 1 should be 10
        matchMaker.execute("matchMaker disconnect_client {\"id\":\"client1\",\"message\":\"job_failed\"}");
        EdgeClientHistory client1History = matchMaker.getClientList().get(0).getClientHistory();
        long thisShouldBe10  = client1History.getHistoryScore("node1");
        Assert.assertEquals(10,thisShouldBe10);
        //wait 2s to see history change ( from 10 score -> 8 score )
        try {
            Thread.sleep(2000);
        } catch (InterruptedException e) {
            e.printStackTrace();
        }
        //Re assign client 1
        matchMaker.execute("matchMaker assign_client client1");
        //2s passed, clint1-node1 score should be 8
        long thisShouldBe8  =  matchMaker.getClientList().get(0).getClientHistory().getHistoryScore("node1");
        Assert.assertEquals(8,thisShouldBe8);

        //Client 1 should connect to node 2 now cuz node 2 have better score
        String thisShouldBeNode2 = getNodeIDFromJSON(matchesList.getMapping().get("client1"));
        Assert.assertEquals("node2",thisShouldBeNode2);
    }
```
Doesn't seem to be working (only the last assert fails), and after debugging it, it looks like it doesnt have anything to do with having added the module . Apparently, the client's score for node2 is always 0, so is not assigning node2 but node1 instead at the end. Could you maybe help me @zero212 ? Given that you understand this way better 😅

alyhasansakr:
@juan-castrillon just comment it for now, we will deal with it later.

juan-castrillon:
Ok, i added an @Ignore tag so we dont forget it

alyhasansakr (changes requested):
Immutable means unchangeable. In other words there should be no way to change the state (instance variables) of the object using its immutable object.

Looking at the implementation here, I can use an ImmutablePlatformModule reference to start and stop a module, that is definitely change of state.

Another case is returning lists, you should return an immutable list not the original list (same for maps). If you just return the list, I can just call the add function of this list.

zero212:
> I did what we discussed last time and implemented the `PlatformObject` class, and made everything I think should extend it do it. However, I am stuck because this test:
> 
> ```java
> @Test
>     public void disconnectAndChangeInScoreTest() throws Exception {
>         matchMaker.execute("matchMaker assign_client client1");
>         //job failed -> client 1 score with node 1 should be 10
>         matchMaker.execute("matchMaker disconnect_client {\"id\":\"client1\",\"message\":\"job_failed\"}");
>         EdgeClientHistory client1History = matchMaker.getClientList().get(0).getClientHistory();
>         long thisShouldBe10  = client1History.getHistoryScore("node1");
>         Assert.assertEquals(10,thisShouldBe10);
>         //wait 2s to see history change ( from 10 score -> 8 score )
>         try {
>             Thread.sleep(2000);
>         } catch (InterruptedException e) {
>             e.printStackTrace();
>         }
>         //Re assign client 1
>         matchMaker.execute("matchMaker assign_client client1");
>         //2s passed, clint1-node1 score should be 8
>         long thisShouldBe8  =  matchMaker.getClientList().get(0).getClientHistory().getHistoryScore("node1");
>         Assert.assertEquals(8,thisShouldBe8);
> 
>         //Client 1 should connect to node 2 now cuz node 2 have better score
>         String thisShouldBeNode2 = getNodeIDFromJSON(matchesList.getMapping().get("client1"));
>         Assert.assertEquals("node2",thisShouldBeNode2);
>     }
> ```
> 
> Doesn't seem to be working (only the last assert fails), and after debugging it, it looks like it doesnt have anything to do with having added the module . Apparently, the client's score for node2 is always 0, so is not assigning node2 but node1 instead at the end. Could you maybe help me @zero212 ? Given that you understand this way better 😅

The reason why the damn client's score never changed (or always 0) is because we are missing 1 POST command in the configuration: `Disconnect` as @alyhasansakr say, just ignore or comment it. I will redo the score based and its tests later

juan-castrillon:
@alyhasansakr You are right, In the last commit I corrected the design.

alyhasansakr (approved):

PR #86, opened 2021-02-05
Physical and network distance calculation

lkh5kor:
Physical distance calculation computation
Get location of a device or application instance implemented 
Modified few getter functions in network class
Network distance draft implementation 

Pending:
network distance computation. 
input and output module functions

### DIFF
--- a/src/main/java/InfrastructureManager/Modules/NetworkStructure/Input/LocationInput.java
+++ b/src/main/java/InfrastructureManager/Modules/NetworkStructure/Input/LocationInput.java
@@ -1,0 +1,26 @@
+package InfrastructureManager.Modules.NetworkStructure.Input;
+
+import InfrastructureManager.ModuleManagement.Exception.Execution.ModuleExecutionException;
+import InfrastructureManager.Modules.NetworkStructure.Exception.NetworkModuleException;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+
+import InfrastructureManager.ModuleManagement.ModuleInput;
+import InfrastructureManager.ModuleManagement.PlatformModule;
+import InfrastructureManager.Modules.NetworkStructure.Network;
+
+public class LocationInput extends ModuleInput {
+	
+	
+	public LocationInput(PlatformModule module,String name,Network network) {
+		super(module,name);
+	}
+    @Override
+    public String read() throws InterruptedException  {
+        String aux = " ";
+        return aux;
+    }
+
+    @Override
+    public void response(ModuleExecutionException outputException) {}
+}
--- a/src/main/java/InfrastructureManager/Modules/NetworkStructure/Network.java
+++ b/src/main/java/InfrastructureManager/Modules/NetworkStructure/Network.java
@@ -69,7 +69,7 @@ public class Network {
 	 * Function to save JSON string
 	 * @return String
 	 */
-	public String saveNetwork() throws InterruptedException{
+	public String saveNetwork(){
 		try {
 			String json = this.mapper.writeValueAsString(this);
 			return json;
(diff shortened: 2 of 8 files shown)


src/main/java/InfrastructureManager/Modules/NetworkStructure/Input/LocationInput.java
```diff
@@ -0,0 +1,39 @@
...
+		super(module,name);
+		this.sharedLocation = this.getSharedLocation();	
+		this.toSend = null;
+	}
+
+	@Override
+	public String read() throws InterruptedException {
+		String aux = "";
+		try {
+			toSend = "location" + sharedLocation.getLocationAsJsonBody();
```
alyhasansakr, line 27: I believe there should be a space after location.

src/main/java/InfrastructureManager/Modules/NetworkStructure/Network.java
```diff
@@ -237,50 +236,222 @@ else if(update == Network.Update.DELETE) {
...
 	/**
-	 * Getter function
-	 * @return applicationDeviceList
+	 * Getter function to get a particular ApplicationInstanceDeviceRelation from applicationDeviceList
+	 * @param id
+	 * @return ApplicationInstanceDeviceRelation
 	 */
-	public List<ApplicationInstanceDeviceRelation> getApplicationDeviceList() {
-		return applicationDeviceList;
+	public ApplicationInstanceDeviceRelation getApplicationInstanceDeviceRelationFromList(String id)	{
```
alyhasansakr, line 243: I don't understand what exactly this function does. I mean what happens if we have an application and a device with the same id?
There are other similar places in this class.

src/main/java/InfrastructureManager/Modules/NetworkStructure/Network.java
```diff
@@ -237,50 +236,222 @@ else if(update == Network.Update.DELETE) {
...
-	 * Getter function
-	 * @return locationConnectionList
+	 * Function to update a particular applicationDevice in applicationDeviceList
+	 * @param applicationDevice
+	 * @param UpdateType // ADD or DELETE
 	 */
-	public List<LocationConnectionRelation> getLocationConnectionList() {
-		return locationConnectionList;
+	public void updateapplicationDeviceList(ApplicationInstanceDeviceRelation applicationDevice,Network.Update update){
+		if (update == Network.Update.ADD)
```
alyhasansakr, line 261: Now is a good time to actually separate the add and delete functions. There should be one function just for adding and another one just for deleting. This applies to the whole class not just this function.

src/main/java/InfrastructureManager/Modules/NetworkStructure/Network.java
```diff
@@ -237,50 +236,222 @@ else if(update == Network.Update.DELETE) {
...
+			}
+		}
+		return idx;
+	}  
+	
+	/**
+	 * Function to get network distance between devices or between application instances.
+	 * @param sourceId and destinationId
+	 */
+	public double getNetworkDistance(String sourceId, String destinationId) {
```
alyhasansakr, line 446: It would have helped me a lot if you had some explanation written of how this is supposed to provide the network distance. I don't understand how it works?

src/main/java/InfrastructureManager/Modules/NetworkStructure/NetworkModuleObject.java
```diff
@@ -2,22 +2,34 @@
...
+	}
 
-    public NetworkModuleObject(ImmutablePlatformModule ownerModule, String name) {
-        super(ownerModule, name);
-    }
 
-    /*
-    TODO: This is an example of a String shared resource, change the type to whatever it is you are sharing,
+	/*
+    TODO: This is an example of a shared Location, change the type to whatever it is you are sharing,
```
alyhasansakr, line 18: Remove the TODO's

src/main/java/InfrastructureManager/Modules/NetworkStructure/Output/DistanceOutput.java
```diff
@@ -0,0 +1,45 @@
...
+	public DistanceOutput(ImmutablePlatformModule module, String name,Network network) {
+		super(module,name);
+		this.network = network;
+	}
+	public void execute(String response) throws NetworkModuleException  {
+		String[] commandLine = response.split(" ");
+		if (commandLine[0].equals("distance")) {
+			try {
+				switch (commandLine[1]) {
+				case "get_physical_distance_devices" -> {
```
alyhasansakr, line 20: It would be better if you have the second word as "physical" and a third word as either "devices" or "application". Also why not locations as well?

src/main/java/InfrastructureManager/Modules/NetworkStructure/Output/DistanceOutput.java
```diff
@@ -0,0 +1,45 @@
...
+					double distance =this.network.getphysicalDistanceBetweenDevices(commandLine[2],commandLine[3]);
+					this.getSharedDistance().putValuetoQueue(distance);
+				}
+				case "get_physical_distance_application" -> {
+					double distance =this.network.getphysicalDistanceBetweenApplications(commandLine[2],commandLine[3]);
+					this.getSharedDistance().putValuetoQueue(distance);
+				}
+				case "get_network_distance" -> {
```
alyhasansakr, line 28: Network distance could also be for devices and applications.

src/main/java/InfrastructureManager/Modules/NetworkStructure/Output/LocationOutput.java
```diff
@@ -0,0 +1,44 @@
...
+	public LocationOutput(ImmutablePlatformModule module, String name,Network network) {
+		super(module,name);
+		this.network = network;
+	}
+	public void execute(String response) throws NetworkModuleException  {
+		String[] commandLine = response.split(" ");
+		if (commandLine[0].equals("location")) {
+			try {
+				switch (commandLine[1]) {
+				case "get_device_location" -> {
```
alyhasansakr, line 22: You already mentioned that it is for a location, here only mention "device"

src/main/java/InfrastructureManager/Modules/NetworkStructure/Shared/SharedDistance.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+
+
+public class SharedDistance extends NetworkModuleObject {
+	private final BlockingQueue distanceQueue;
+
+	public SharedDistance(ImmutablePlatformModule ownerModule) {
+		super(ownerModule);
+		this.distanceQueue = new ArrayBlockingQueue<>(25);
```
alyhasansakr, line 15: I believe ArrayBlockingQueue is buffer bound, which means it has a limited size (25 here), why don't you use an unlimited queue?
Also why don't you define the type? (it is double).

alyhasansakr (changes requested):
Remove all the unnecessary spaces and tabs.
In general I dislike the way we implement the getters for the Network class. You should be able to index through the lists rather than doing a linear search every time a piece of information is needed. Each instance (device, location and so on) must have a unique ID and thus could be indexed (normally using a HashMap).

PR #88, opened 2021-02-09
Fixed input and output names in match making modules

### LINKED ISSUE #87: Error when run the application
Current master branch return error when runs application (no files were modified)

Here is the screenshot.

image

Am I being dump for not knowing something again or is it really not working...Would you mind have a look at it @juan-castrillon? 

Thanks buddy <3

juan-castrillon:
Request to fix #87 .Inputs and outputs in the new Match Making modules, were not name correctly so creating connections to it threw exceptions.


zero212 (approved):
LGTM. Things I needed when through. Thanks again xD

PR #91, opened 2021-02-10
Hot fix shared resources

juan-castrillon:
Based on the discussion, this is the first attempt to refactor the resource sharing in modules.

So far:
- `ImmutablePlatformModule` is now an interface
- Added a `GlobalVarAccessPlatformModule` interface to represent limited access (only access to the global variables)
- In the super class `PlatformObject`, added a method to get the owner module with global variable access
- Added a `ModuleShareResource` interface, which the different shared objects in the modules will implement
- In the modules, there is now a `globalVariables` map that relates a name to a `ModuleSharedResource`
- Modified `RemoteExecutionModule` to work with the new functionalities and tests it still works

The biggest issue is that is order to have a `ModuleShareResource` interface (or even to have a `GlobalVarAccessPlatformModule` interface) the shared resources have to be abstracted somehow. So in this first iteration, a shared resource is characterized by having a method to modify it (which recieves data as Strings). I realize that is very general (and may be wrong) but the alternative would be to use generics, and either making everything generic (`PlatformModule` generic) or just making the `GlobalVarAccessPlatformModule` interface generic but then, each module should implement it separately (the abstract `PlatformModule` could not implement it, not knowing the type of the shared resource).

Have a look at the code, and let me know what you think @alyhasansakr

### DIFF
--- a/src/main/java/InfrastructureManager/ModuleManagement/GlobalVarAccessPlatformModule.java
+++ b/src/main/java/InfrastructureManager/ModuleManagement/GlobalVarAccessPlatformModule.java
@@ -1,0 +1,5 @@
+package InfrastructureManager.ModuleManagement;
+
+public interface GlobalVarAccessPlatformModule {
+    public ModuleSharedResource getResource(String resourceName);
+}
--- a/src/main/java/InfrastructureManager/ModuleManagement/GlobalVarAccessPlatformModule.java
+++ b/src/main/java/InfrastructureManager/ModuleManagement/GlobalVarAccessPlatformModule.java
@@ -1,5 +1,0 @@
-package InfrastructureManager.ModuleManagement;
-
-public interface GlobalVarAccessPlatformModule {
-    public ModuleSharedResource getResource(String resourceName);
-}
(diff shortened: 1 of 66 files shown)


alyhasansakr:
@juan-castrillon Why not have a module specific object class (NetworkObject for example) which extends PlatformObject. This way you can implement global access functions directly on each module specific object class (no need for GlobalVarAccessPlatformModule interface or similar).

juan-castrillon:
The last commit is implementing the latest idea:
- Input and output are now interfaces
- Platform obeject now has the fields they use to have (name basically)
- Runners are now kept by a map in the module and the input does not know about it
- Each module has a particular class that implements PlatformObject and allows for global var access
- It also has another "access" interface to cast the module into in this new object class (That way we dont access anything apart from the global variables)
- The whole thing will change all modules, however just for discussion, I only commited the changes done in RemoteExecution module so we can have a look

alyhasansakr:
@juan-castrillon That looks like a good idea, go on with the rest.

juan-castrillon:
In the last commits:

- I've refactored all modules to have a module specific implementation of `PlatformObject` and for their inputs and outputs (and all other elements) to implement it
- In modules where shared resources are used, I've added a "intermediate" access interface and provided access to the shared resources via the local `PlatformObject` implementation
- Fixed all testing to reflect the new changes
- Added a "dummy" implementation in the `NetworkModule` with a String as a shared resource (to be changed later to what it needs to be shared). This includes the local object, and the access interface

alyhasansakr (approved):

PR #92, opened 2021-02-18
Javadocs v1

### LINKED ISSUE #80: Documentation for Modules Refactor
Because of the new modules refactor. merged with #78. The following documentation needs to be added:

- The wiki here in github, needs to reflect the new way to configure and use the platform
- Each module should have its own wiki page (reference and example)
- A minimal configuration for each module should be included as well (just connecting the module to a console).
-  Code documentation (javadoc) is required for all code in master

juan-castrillon:
Pull request contaning code documentation to close #80. 

The current last commit contains code documented in the
- Configuration package
- ModuleManagement package
- PlatformObject class
- Master class

Missing are documenting each specific module.

Also, I have the files ready to make the javadoc deployment automatic using GH-Pages, GH-actions and gradle. However, because it requires to activate the GH pages deployment, I will need access to the settings of the repo or you can enable the deployment @alyhasansakr.


alyhasansakr:
@juan-castrillon It should be generated form this branch using the folder /docs
After we merge, I will change it back to the master branch

juan-castrillon:
I tested the current workflow in another repository of mine and it should work. The action is copying the files into the GH worker, then generating javadocs (in a folder called `/javadocs`). Then, it grabs everything in that folder and deploys it to the branch `javadocs_v1` into the `docs` folder. Finally. github pages should pick up that folder and put it in a webpage

Like i said, i tried the exact same thing in another [repo](https://github.com/juan-castrillon/ActionsTest) and it works, but here I keep getting this error:

_No /docs folder was found to build GitHub Pages. Check the source setting for this repository._

And checking this branch, no docs folder has been added indeed. 

I believe github pages is trying to deploy (automatically since you activate it), but the action is not running (because is only triggered by pushes to master, as I think makes sense) so of course no docs folder is created. 

I can think of different options to solve it:

-  I run `javadocs` locally once, create the `docs` folder and push, so github pages has something to deploy. This will later be overriden when new pushes to master are made.
- Deploying the page from a different branch. We could have a dedicated branch, Something like `gh-pages`  or `doc-page` or something similar and we can deploy the github page from root there (no docs folder). That way github pages is happy for now, although it deploys an empty branch. Which will be filled once push to master occurs.
- You turn off github pages and only turn it on after we merge this. That way the action will already run

@alyhasansakr  what do you think?

alyhasansakr:
@juan-castrillon I changed it to be in the master under /docs.
I can merge it even if the test fails with the other account.

juan-castrillon:
Missing are only:
-REST Module
- Scenario Module

I believe NetworkModule and MatchMaking should be documented by shankar and zero respectively when they finish them

juan-castrillon:
All packages are commented, I believe is ready to be merged

Issue #97, opened 2021-03-01
End-to-End Latency Report Runner


theodorelu27:
Implement a new component to the Client Side of the application, in which the main role was to submit the end-to-end latency of each iterations to the Master through REST

Components to be added to the Client:

- A new thread runner that would wait for input from the main operator
- A new operator that would handle the send, with REST, the following information: End-to-End Latency, Server Usage, Server ID, timestamp

Components to be added to the Master:

- REST input for the information
- Module to receive the input, process, and output into format that can be processed in MathLab

alyhasansakr:
@theodorelu27 I agree on the client part, but I believe the master part should be already possible with what we currently have. That is to receive a REST input (as JSON) and write the value to a file with the file writer module. This file can then be processed in MatLab.

theodorelu27:
@alyhasansakr ah okay, thanks for confirming
I'd update the requirement to "REST Input" and "Filewriter output", these two would be implemented as a connections pair in the configuration file

PR #101, opened 2021-03-11
Implementation for Latency Report Runner

theodorelu27:
Implementation for issue #97 

High-Level Contents:
- Added new operator and runner that would send latency report to Master every 1 second through rest
- Configuration for Latency Reporting

Up to this point:
- Report is generated in .txt and one report file would be created for each client
- new report would be added without delimiters

sample:
`{"timestamp":"2021-57-11_20:57:13","latency":148,"use_server":false,"server_ip":null}{"timestamp":"2021-57-11_20:57:17","latency":2966,"use_server":true,"server_ip":"172.17.0.3"}{"timestamp":"2021-57-11_20:57:46","latency":664,"use_server":false,"server_ip":"172.17.0.3"}`

so there were 3 conditions that could be inferred from the sample:

1.  Image was processed in the Client before connecting to the Server
`{"timestamp":"2021-57-11_20:57:13","latency":148,"use_server":false,"server_ip":null}`
2. Image was processed in the Server
`{"timestamp":"2021-57-11_20:57:17","latency":2966,"use_server":true,"server_ip":"172.17.0.3"}`
3. Image was processed in the Client because of connection failure to the Server
`{"timestamp":"2021-57-11_20:57:46","latency":664,"use_server":false,"server_ip":"172.17.0.3"}`

@alyhasansakr are there any adjustments that need to be made to make this information processable in MatLab?

Also, I'm not sure who to ask, but after I assigned a node with `assign_client`, I received `ClientAlreadyAssignedException` in response to the register_client that is currently used for the client's Heartbeat, would I need to change this?


alyhasansakr:
@theodorelu27 Looks good to me.
About the exception, you should ask @zero212

zero212:
assigned a node? 
The way the match making work is you assign a client to a node.
Let say we have `client1` and `client2`, `node1`
That exception `ClientAlreadyAssignedException` only appear, if you already have `client1` assigned to `node1` that client1, and 
try to assign `client1` to `node1` again. 

In short:
You can only assign `client2` to `node1` at this time. Assigning `client1` to `node1` will result in that exception. 
You should not change the code, but change the approach to assigning client.
There are actually a function to disconnect client to node so you can reassign client1 to node1, but that is not implemented in REST (yet)

theodorelu27:
@alyhasansakr thank you
@zero212 I received the error not when I was trying to assign the node for the second time (the node assignment is currently done manually through the console, REST request, or scenario)
the error was received when the register_client was invoked through REST request that the Client application ran as its Heartbeat signal. Thus, if register_client is no longer useable in this context, then I would need to update the code on the Client application.

Also, this error started to appear as soon as I assigned the node for the first time, here's the stacktrace
`InfrastructureManager.Modules.MatchMaking.Exception.ClientAlreadyAssignedException: can't update because [client13] already been assigned to [{"id":"13","ipAddress":"172.17.0.3:10300","connected":false,"resource":0,"network":0,"location":0,"totalResource":0,"totalNetwork":0}]
	at InfrastructureManager.Modules.MatchMaking.Output.MatchMakerOutput.registerClient(MatchMakerOutput.java:147)
	at InfrastructureManager.Modules.MatchMaking.Output.MatchMakerOutput.execute(MatchMakerOutput.java:388)
	at InfrastructureManager.ModuleManagement.PlatformModule.lambda$new$0(PlatformModule.java:76)
	at InfrastructureManager.ModuleManagement.Runner.run(Runner.java:74)
	at java.base/java.lang.Thread.run(Thread.java:832)
`

and thank you for the information regarding the disconnection, I might need to do so for my experiment so that was good to know

alyhasansakr:
@juan-castrillon This is the same issue you asked about today.
@zero212 @theodorelu27 From my point of view, if a heart beat signal is not received for a set period from both the client and the server, then the master should automatically drop the registration (and consider them disconnected).

zero212:
#102 I create the issue for heartbeat signal. Please have a look and chime in your thoughts or help. 
Love love <3 
@theodorelu27 @juan-castrillon @alyhasansakr

theodorelu27:
@alyhasansakr okay so for my part:
- separate registration from heartbeat and get server
- heartbeat should be on a fixed interval (configurable)
- so after registration -> start heartbeat and get server
- if connected client is offline, server must say in its heartbeat that it's no longer connected

theodorelu27:
@alyhasansakr I've applied the following changes:
- separate registration from heartbeat and get server
- heartbeat should be on a fixed interval (configurable)
- so after registration -> start heartbeat and get server
- if connected client is offline, server must say in its heartbeat that it's no longer connected
so after @zero212 is done with the matchmaking part, I should be able to test after making some adjustments to the configuration file

theodorelu27:
@alyhasansakr I've implemented the registration and heartbeat, probably need to confirm some stuff:
- Do we still need the server to send its connected status to the master as currently, the matchmaking isn't using it?
- I think I misunderstood the value for `interval` that we have in our last discussion, the current implementation is based on the definition that it is `the time interval until the edge application sends another heartbeat to the master` while what I understood from Zero's explanation was more into `the time interval for master to wait before it declares that the edge application is offline` 

Remaining tasks
- [ ] Check on timestamps, as currently the timestamps could be way off and there were some instances where the month value is either 0 or is more than 12
- [ ] add TODOs for currently hardcoded values such as reqNetwork, reqResource, and location

theodorelu27:
@alyhasansakr turns out I just wrote the wrong date format, using minute as the month, the other incorrect value was due to timezone difference, time is still consistent so for experimentation, It might not impact the result

so aside from the stuffs I wanted to confirm in the previous comment, should be all complete for this issue

alyhasansakr:
@theodorelu27 

- For now, just send any value for connected (we will rewrite later).
- Ideally both should be the same. Practically, it is advised to send the heartbeat at twice as much rate as the deadline. The implementation done by Zero is the deadline, which means that your application needs to send a heartbeat message two times during one interval. For example, if you send the interval to the master as 10 seconds, then send the heartbeat every 5 seconds.
- These should be parameters (environment variables for Docker).

zero212:
- Send true for connected, matchmaking won't work if connected = false

theodorelu27:
@alyhasansakr I see, currently, as a buffer I doubled the interval that I sent to the Master as a deadline, so with 1000ms interval, the deadline is 2000ms
I'd try to set it to be the same and see how it goes first

theodorelu27:
@alyhasansakr I tried to submit the interval = deadline and all signals were on time, I think we can keep it like this for now

alyhasansakr:
@theodorelu27 That is because you don't have any latency in your setup, once you do it will not arrive in time anymore.

alyhasansakr (approved):

Issue #103, opened 2021-03-13
Master as a Docker container


alyhasansakr:
It should be possible to run the main master application as a Docker container.

We need to do the following:

1. A Gradle task that creates the image and pushes it to a local Docker registry (hosted on a local machine) or directly to a local Docker cache (same as if you build an image directly on your local machine).
2. A GitHub action that creates the image and pushes it to a public Docker registry (when changes are pushed to the master branch).

Of course we still need the original Gradle run task that just runs the application directly.

juan-castrillon:
After doing some research in avaiable plugins for this I think i will follow this structure for the gradle part:
- Create a `fatJar` task (Same as the one in client and server of my app) for the platform
- Manually define a dockerfile similar to the one in the app (Multistage build with gradle image and jre)
- Create a `buildImage` task that uses one of the available gradle plugins (either [this](https://bmuschko.github.io/gradle-docker-plugin) or [this](https://github.com/palantir/gradle-docker)) to create the docker image and have it in local cache
- Create a `dockerPushLocal` task that pushes to a locallly deployed repo (Specifying configuration in the `gradle.build` file)
- Create a `dockerPushRemote` task that pushes to a dockerhub repo (Specifying configuration in the `gradle.build` file)

I dont think we need a `dockerRunContainer` task or something similar.

Additionally, for the dockerhub part i will initially try it with a dummy account

alyhasansakr:
OK, that should be good enough.
One more point to consider in case someone uses only the Docker image, how should the configuration be passed to a Docker container?

juan-castrillon:
I am struggling with something related right now. The way we load the configuration file in the master makes it dependant on the path where java is executed (actually the way we load all files, e.g scenarios as well). So when creating the jar, running from the main directory worked but going to where the jar is and running there didnt. I managed to change that so that the jar reads from the default configuration file that is bundled inside as a resource. However, this makes it so a custom configuration file needs to be in the `resources` folder that is generated when the project is built, so a configuration file outside cannot be read (that would be an issue if we copy the config file in the dockerfile for example and try to run the platform passing the path as argument)

[...]

PR #104, opened 2021-03-17
Docker deployment

juan-castrillon:
Pull request to solve #103 

### New Tasks
As of right now, the following is added to the gradle project:
- A `fatJar` task that creates a  jar file for the platform that includes all the dependencies.
- A `docker` task that based on the following dockerfile, builds a docker image and passes the generated jar (This one calls `fatJar` automatically)

    ```dockerfile
    #Docker file to create images from fatJar custom task
    FROM adoptopenjdk/openjdk15:jre-15.0.2_7-alpine
    EXPOSE 4567
    RUN mkdir /EdgeDiagnostics
    #copy the generated jar into a lightweight image with only JRE
    COPY *.jar /EdgeDiagnostics/platform.jar
    WORKDIR /EdgeDiagnostics
    ENTRYPOINT ["java", "-jar","./platform.jar"]
    #Use this to put arguments that pass as default to the app, otherwise pass them in docker build
    CMD []
    ```
- Configurable `dockerPush` tasks. For each of the declared repositories in the `tag` section in the gradle build file, a task is created. In the current commit, a repository in my docker hub account is used and named `JuanDockerHub`. Calling the task `gradle dockerPushJuanDockerHub`  then pushes the image to the repository (Requires previous login)

### Running the docker image
With respect to working with the image:

To run the image, different parameters should be used according to what is required.

```
docker run -it -p 4567:4567 --name <container_name> edge_diagnostic_platform:0.5
```
Runs the platform with the default configuration (`src/main/resources/Configuration.json`)

```
docker run -it -p 4567:4567 -v C:\Users\jpcr3\Desktop\Projects\FH-Dortmund\Edge-Diagnostic-Platform\Scenarios:/EdgeDiagnostics/Scenarios --name diagnostics edge-diagnostic-platform:0.5
```
Runs the platform with the default configuration and passes a folder in which scenarios are included, so
the platform can read them. In this case, the path for the scenarios shoulb be specified in the config file in the container context (Is easier if the two folders are called the same)

```
docker run -it -p 4567:4567 -v C:\Users\jpcr3\Desktop\Projects\FH-Dortmund\Edge-Diagnostic-Platform\Scenarios:/EdgeDiagnostics/Scenarios -v C:\Users\jpcr3\Desktop\Projects\FH-Dortmund\Edge-Diagnostic-Platform\Configurations:/EdgeDiagnostics/Configurations --name <container_name> edge-diagnostic-platform:0.5 -c Configurations/UtilityModuleTestConfiguration.json
```
Runs the platform, and mounts a folder for Scenarios and one for Configurations. The argument `Configurations/UtilityModuleTestConfiguration.json`
tells the platform to start with the custom configuration located in the mounted folder.

Another folder can also be passed to store files generated by a `fileOutput` for example. (Mounting a volume is biderectional)

### Missing

Still missing are:
- Defining repositories to which we want to push (Maybe create a FH-Edge-Lab docker hub account, or setting up a self-hosted repo in the university)
- Getting credentials for the repos (Username and Password)
- Github Actions to perform the process on pushes to master (Also requires credentials to the repos)

@alyhasansakr What do you think about the repos?

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,40 @@
+name: Build and push docker image
+on:
+  push:
+    branches:
+      - master
+jobs:
+  deploy-on-docker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+      - name: Set up JDK 14
+        uses: actions/setup-java@v1
+        with:
+          java-version: 14
+      - name: Cache Gradle packages
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+      - name: Grant execute permission for gradlew
+        run: chmod +x gradlew
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+      - name: Build jar, build image and push with gradle
+        run: ./gradlew dockerPushDockerHub
+      - name: Cleanup Gradle Cache
+          # Remove some files from the Gradle cache, so they aren't cached by GitHub Actions.
+          # Restoring these files from a GitHub Actions cache might cause problems for future builds.
+        run: |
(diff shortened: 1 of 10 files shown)


alyhasansakr:
@juan-castrillon Create a docker hub account for FH-EDGE-LAB and push the image to it on master push using GitHub actions.

juan-castrillon:
The user `edgelabfh` is now created in [DockerHub](https://hub.docker.com/u/edgelabfh). I pushed the image manually by using only the task `gradle dockerPushDockerHub`. This builds the uberJar, builds the image, tags it and pushes it.

A github actions workflow was also added which does the exact same thing. Logs into docker and then calls the `dockerPushDockerHub` task

I believe is ready for merging

alyhasansakr (approved):

PR #115, opened 2021-04-14
Implement EQR Evaluation

### LINKED ISSUE #98: Edge Quality Rating Evaluation
Implement Edge Quality Rating for the Client's Connection Evaluator

Components to be added to the Clients:

**Operator for EQR calculation**
- Counter for current EQR score
- Configureable Latency Treshold and score reduction
- Include it as information to be sent on the Latency Reporter ( #97 )

theodorelu27:
Implementation for #98 with fixed max EQR, evaluating state, and disconnect client when EQR is not satisfactory


theodorelu27:
@alyhasansakr still a mess, but all planned features has been implemented

src/main/java/Application/MarkerDetection/OpenCVClient/ServerEvaluationOperator.java
```diff
@@ -0,0 +1,34 @@
...
+        activeOperator.markerDetection(true, true, detector);
+
+        long endTime = System.nanoTime();
+        long latency = (endTime-startTime)/1000000;
+
+        if(latency > latencyThreshold) {
+            evaluator.decrementEQR();
+        } else {
+            evaluator.incrementEQR();
+        }
```
alyhasansakr, line 31: take this condition out in a separate thread. Use the last latency value (you will need a Semaphore)

theodorelu27:
@alyhasansakr I've implemented the EQR Change queueing and management runner
but I just realized that after disconnecting, while disconnect seems to work as intended, the get node would still provide the last provided IP, so in the client side, it would try to connect and ended up with a bad evaluation and another disconnect

alyhasansakr (changes requested):
I still don't see how the rate of increments and decrements is fixed. What you did here is to fix the rate at which they are applied, but not at which they are generated. Still after your solution, the higher the latency, the slower the change rate of EQR ("enqueue" rate not "manage" rate).

theodorelu27:
@alyhasansakr yeah, I just realized that it would be applied quite differently from what I initially expected
I've changed the EQR increment generation to a fixed time according to the last recorded latency as we discussed
but as EQR is reported using the latency reporter, the faster change of the EQR (now updated every 10ms) will not be reflected on smaller EQR, is that okay?

```
{"timestamp":"2021-04-21_11:43:44","latency":122,"use_server":false,"server_ip":null,"evaluating":false,"eqr":0}
{"timestamp":"2021-04-21_11:43:45","latency":93,"use_server":false,"server_ip":null,"evaluating":false,"eqr":0}
{"timestamp":"2021-04-21_11:43:46","latency":197,"use_server":false,"server_ip":"172.17.0.3","evaluating":true,"eqr":5}
{"timestamp":"2021-04-21_11:43:48","latency":102,"use_server":false,"server_ip":"172.17.0.3","evaluating":true,"eqr":10}
{"timestamp":"2021-04-21_11:43:49","latency":448,"use_server":true,"server_ip":"172.17.0.3","evaluating":false,"eqr":10}
{"timestamp":"2021-04-21_11:43:51","latency":513,"use_server":true,"server_ip":"172.17.0.3","evaluating":false,"eqr":10}
```

alyhasansakr (dismissed):

alyhasansakr:
@theodorelu27 That is ok, actually the speed at which the EQR updates is inversely proportional to the max EQR. Now that we change it every 10ms we should have a higher max EQR. For that max EQR should be configurable.

alyhasansakr (dismissed):

